PR #15, opened 2020-09-04
Adaptdl registry

### LINKED ISSUE #13: AdaptDL submit with insecure registry doesn't work when scheduler is in a non-default namespace
Steps to reproduce:

1. Install AdaptDL scheduler to the `adaptdl` namespace with `helm install adaptdl adaptdl --repo https://github.com/petuum/adaptdl/raw/helm-repo   --namespace adaptdl --create-namespace --set docker-registry.enabled=true`
2. Submit a job in the `default` namespace using `adaptdl submit ...`

Result:

- `kubectl get pods` shows `myjob-6a5a039c-adea-4eb2-bdd1-7639d9893e63-0-0   0/1     ErrImagePull   0          12s`
- `kubectl describe pod` shows an error:
```
Failed to pull image "adaptdl-registry.default.svc.cluster.local:31001/dev/adaptdl-submit@sha256:988db89a3ce7dbc6f675b77fafef36e23704c8dd8621d6121450ff51261bf086": rpc error: code = Unknown desc = failed to resolve image "adaptdl-registry.default.svc.cluster.local:31001/dev/adaptdl-submit@sha256:988db89a3ce7dbc6f675b77fafef36e23704c8dd8621d6121450ff51261bf086": no available registry endpoint: failed to do request: Head "https://adaptdl-registry.default.svc.cluster.local:31001/v2/dev/adaptdl-submit/manifests/sha256:988db89a3ce7dbc6f675b77fafef36e23704c8dd8621d6121450ff51261bf086": http: server gave HTTP response to HTTPS client
```

### LINKED ISSUE #14: AdaptDL submit with insecure registry doesn't work with MicroK8s
Steps to reproduce:

1. Deploy MicroK8s following https://adaptdl.readthedocs.io/en/latest/installation/deploy-microk8s.html
2. Install AdaptDL scheduler to the `default` namespace using `helm install adaptdl adaptdl --repo https://github.com/petuum/adaptdl/raw/helm-repo --namespace default --create-namespace --set docker-registry.enabled=true`
3. Submit a job using `adaptdl submit ...`

Result:

- `kubectl get pods` shows `my-job-9gfl9-47b5f562-b96d-443a-a654-e7a6c27a66fc-0-0   0/1     ImagePullBackOff   0          44s`
- `kubectl describe pod` shows an error:
```
Failed to pull image "adaptdl-registry.default.svc.cluster.local:31001/dev/adaptdl-submit@sha256:988db89a3ce7dbc6f675b77fafef36e23704c8dd8621d6121450ff51261bf086": rpc error: code = Unknown desc = failed to resolve image "adaptdl-registry.default.svc.cluster.local:31001/dev/adaptdl-submit@sha256:988db89a3ce7dbc6f675b77fafef36e23704c8dd8621d6121450ff51261bf086": no available registry endpoint: failed to do request: Head "https://adaptdl-registry.default.svc.cluster.local:31001/v2/dev/adaptdl-submit/manifests/sha256:988db89a3ce7dbc6f675b77fafef36e23704c8dd8621d6121450ff51261bf086": http: server gave HTTP response to HTTPS client
```

odp:
Issues handled by this PR

1. We now use port `32000` for registry service everywhere to gain compatibility with microk8s (which is its default) 

2. For internal registry we patch reponame of the ADL job with `localhost`+suffix so that image can be pulled on the remote hosts without need for any `/etc/hosts` entry change on remote host node. This is applicable to both microk8s and EKS.

3. Handle stale `/etc/hosts` entries.

4. Handle case where registry is deployed in a different namespace 

Fixes #13 
Fixes #14

### DIFF
--- a/adaptdl_cli/adaptdl_cli/registry.py
+++ b/adaptdl_cli/adaptdl_cli/registry.py
@@ -54,11 +54,17 @@ def _get_node_ip():
 def fix_etc_hosts():
     # Correct /etc/hosts entry with the current node IP kubectl is pointing to
     node_ip = _get_node_ip()
+    host_entry = None
     if not node_ip:
         raise SystemExit("Didn't find any nodes.")
-    if node_ip != socket.gethostbyname(ADAPTDL_REGISTRY_URL.split(':')[0]):
+    try:
+        host_entry = socket.gethostbyname(ADAPTDL_REGISTRY_URL.split(':')[0])
+    except socket.gaierror:
+        pass
+
+    if node_ip != host_entry:
         assert os.system(f"sudo `which hostman` add -f {node_ip} \
-                          {ADAPTDL_REGISTRY_URL.split(':')[0]}") == 0
+                {ADAPTDL_REGISTRY_URL.split(':')[0]}") == 0
 
 
 def _find_entry():
--- a/adaptdl_cli/adaptdl_cli/registry.py
+++ b/adaptdl_cli/adaptdl_cli/registry.py
@@ -16,16 +16,21 @@
 
 import json
 import os
-import subprocess
 from pathlib import Path
 import sys
 import socket
+import kubernetes
+
 
 ADAPTDL_REGISTRY_URL = "adaptdl-registry.remote:32000"
-ADAPTDL_REGISTRY_CREDS = "adaptdl-registry-creds"
 
-_DAEMON_FILE = f"{str(Path.home())}/.docker/daemon.json"
+
+if "linux" in sys.platform.lower():
+    _DAEMON_FILE = "/etc/docker/daemon.json"
+else:
+    _DAEMON_FILE = f"{str(Path.home())}/.docker/daemon.json"
 _REG_KEY = "insecure-registries"
+
 
 # https://forums.docker.com/t/restart-docker-from-command-line/9420/9
 MACOS_DOCKER_RESTART_SCRIPT = r"""osascript -e 'quit app "Docker"'; \
@@ -37,17 +42,25 @@ LINUX_DOCKER_RESTART_SCRIPT = r"""sudo systemctl restart docker \
 
 
 def _get_node_ip():
-    nodes = json.loads(subprocess.check_output(['kubectl', 'get', 'no',
-                                                '--sort-by=.metadata.creationTimestamp',  # noqa: E501
-                                                '-ojson']))
-    ip = None
-    if len(nodes['items']) > 0:
-        for addr in nodes['items'][0]['status']['addresses']:
-            if addr['type'] == 'InternalIP':
-                ip = addr['address']
-            if addr['type'] == 'ExternalIP':
-                ip = addr['address']
-                break
+    def ready(conditions):
+        return any(cond.type == "Ready" and
+                   cond.status == "True" for cond in conditions)
+
+    # Collect Ready nodes sorted by descending age
+    v1 = kubernetes.client.CoreV1Api()
+    nodes = v1.list_node()
+    nodes = [node for node in nodes.items if ready(node.status.conditions)]
+    nodes = sorted(nodes, key=lambda x: x.metadata.creation_timestamp)
+    if len(nodes) == 0:
+        return None
+
+    for addr in nodes[0].status.addresses:
+        if addr.type == 'InternalIP':
+            ip = addr.address
+        if addr.type == 'ExternalIP':
+            ip = addr.address
+            break
+    assert ip
     return ip
 
 
@@ -93,11 +106,10 @@ def _fix_daemon_json():
 
 
 def registry_running():
-    svc = json.loads(subprocess.check_output(['kubectl', 'get', 'svc',
-                                              '-A', '-l',
-                                              'app=docker-registry',
-                                              '-ojson']))
-    return len(svc['items']) > 0
+    v1 = kubernetes.client.CoreV1Api()
+    svc = v1.list_service_for_all_namespaces(
+        label_selector="app=docker-registry,release=adaptdl")
+    return len(svc.items) > 0
 
 
 def fix_local_docker():
--- a/adaptdl_cli/adaptdl_cli/registry.py
+++ b/adaptdl_cli/adaptdl_cli/registry.py
@@ -14,16 +14,16 @@
 # limitations under the License.
 
 
-import json
-import os
-from pathlib import Path
 import sys
+import os
 import socket
+import json
+import subprocess
+from pathlib import Path
 import kubernetes
 
 
 ADAPTDL_REGISTRY_URL = "adaptdl-registry.remote:32000"
-
 
 if "linux" in sys.platform.lower():
     _DAEMON_FILE = "/etc/docker/daemon.json"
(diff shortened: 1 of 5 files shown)


adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -75,15 +93,14 @@ def _fix_daemon_json():
 
 
 def registry_running():
-    svc = json.loads(subprocess.check_output(['kubectl', 'get', 'svc', '-l',
-                                             'app=docker-registry', '-ojson']))
+    svc = json.loads(subprocess.check_output(['kubectl', 'get', 'svc',
+                                              '-A', '-l',
+                                              'app=docker-registry',
+                                              '-ojson']))
```
aurickq, line 99: Possible to do a more targeted/specific query by adding an annotation to the registry adaptdl uses? i.e https://github.com/helm/charts/blob/master/stable/docker-registry/values.yaml#L29

adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -37,6 +38,7 @@
 
 def _get_node_ip():
     nodes = json.loads(subprocess.check_output(['kubectl', 'get', 'no',
+                                                '--sort-by=.metadata.creationTimestamp',  # noqa: E501
```
aurickq, line 41: Don't circumvent linter

adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -75,15 +93,14 @@ def _fix_daemon_json():
 
 
 def registry_running():
-    svc = json.loads(subprocess.check_output(['kubectl', 'get', 'svc', '-l',
-                                             'app=docker-registry', '-ojson']))
+    svc = json.loads(subprocess.check_output(['kubectl', 'get', 'svc',
```
aurickq, line 96: Prefer to use the `kubernetes` dependency package instead of calling `kubectl` in a subprocess. This is the convention elsewhere in the CLI, since we have control over the `kubernetes` package version but not the `kubectl` version.

adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -49,6 +51,22 @@ def _get_node_ip():
...
+    host_entry = None
+    if not node_ip:
+        raise SystemExit("Didn't find any nodes.")
+    try:
+        host_entry = socket.gethostbyname(ADAPTDL_REGISTRY_URL.split(':')[0])
+    except socket.gaierror:
+        pass
+
+    if node_ip != host_entry:
+        assert os.system(f"sudo `which hostman` add -f {node_ip} \
```
aurickq, line 66: Why `os.system` and not `subprocess.run`?

adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -49,6 +51,22 @@ def _get_node_ip():
...
+                {ADAPTDL_REGISTRY_URL.split(':')[0]}") == 0
```
aurickq, line 67: If nodes change, will this command append to /etc/hosts or replace the existing entry?

adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -19,8 +19,9 @@
 import subprocess
 from pathlib import Path
 import sys
+import socket
 
-ADAPTDL_REGISTRY_URL = "adaptdl-registry.default.svc.cluster.local:31001"
+ADAPTDL_REGISTRY_URL = "adaptdl-registry.remote:32000"
 ADAPTDL_REGISTRY_CREDS = "adaptdl-registry-creds"
```
aurickq, line 25: Do we need these cred anymore?

adaptdl_cli/adaptdl_cli/registry.py
odp: yes

adaptdl_cli/adaptdl_cli/registry.py
odp: Sure. However I found `kubectl` usage at multiple places in the cli e.g in `cp`, `logs`, `ls` etc. We are entirely not independent of `kubectl`.

adaptdl_cli/adaptdl_cli/registry.py
odp: replace

adaptdl_cli/adaptdl_cli/registry.py
odp: not really

adaptdl_cli/adaptdl_cli/registry.py
odp: because it's simple. Also I'm not interested in reading out/err etc.

aurickq: In that case, using `subprocess.run` is preferable. It's the way recommended by Python, and nearly as simple. Also it will automatically throw an exception instead of having to use `assert`. The entire statement in the `assert` can be disabled via an environment variable, which we don't have control over.

adaptdl_cli/adaptdl_cli/registry.py
```diff
@@ -75,15 +106,13 @@ def _fix_daemon_json():
 
 
 def registry_running():
-    svc = json.loads(subprocess.check_output(['kubectl', 'get', 'svc', '-l',
-                                             'app=docker-registry', '-ojson']))
-    return len(svc['items']) > 0
+    v1 = kubernetes.client.CoreV1Api()
+    svc = v1.list_service_for_all_namespaces(
+        label_selector="app=docker-registry,release=adaptdl")
```
aurickq, line 111: I don't think this hard-coded selector works in general since users can deploy AdaptDL under a different release name, for example if they install the chart using `-g` instead of exactly specifying the release name as `adaptdl`

odp: We assign a fixed name `adaptdl-registry` to the service. So `field_selector` should work in any deployment scenario.

adaptdl_cli/adaptdl_cli/registry.py
odp: `subprocess.run` takes arguments as list of strings. I'm running small scripts as part of daemon restart which is hard to break down and makes the code/command unreadable. Besides this is completely unrelated to current PR and the problem it fixes.

aurickq: I think you can just pass "shell=True" into `subprocess.run` and it will accept a command as a string. From python docs:
```
>>> subprocess.run("exit 1", shell=True, check=True)
Traceback (most recent call last):
  ...
subprocess.CalledProcessError: Command 'exit 1' returned non-zero exit status 1
```

adaptdl_cli/adaptdl_cli/registry.py
aurickq: SGTM. Can you also add a comment to `fullnameOverride` in `values.yaml` saying it should not be changed and why?

adaptdl_cli/adaptdl_cli/registry.py
odp: that works. thanks.

adaptdl_cli/adaptdl_cli/registry.py
odp: Added the comment but I think it's kind of implied since someone used name override in the first place.

aurickq (approved):

PR #26, opened 2020-09-14
Adds gradient accumulation

rmfan:
- Adds an option to `trainloader.autoscale_batch_size` to allow for gradient accumulation
- Allows for the adaptdl scheduler to allocate logical local batch sizes up to `grad_acc_steps` times larger than the provided batch size bound. If it is larger than the provided bounds, then gradient accumulation is used
- Modify speedup modeling to account for gradient accumulation
- If gradient accumulation is used, then the extra steps are used as additional datapoints for AdascaleSGD
- Modifies the lr scheduler's step function to be a no-op during an intermediate step of gradient accumulation
- Adds a new `get_progress` function to AdascaleSGD to replace the existing gain function to determine how much progress is made in AdascaleSGD's scale-invariant iteration. Notably, it returns zero during and intermediate step of gradient accumulation. The `gain` property remains unchanged.


adaptdl/adaptdl/speedup.py
```diff
@@ -53,15 +53,23 @@ class SpeedupFunction(object):
 
     def __init__(self, params, grad_params=None, init_batch_size=None,
                  max_batch_size=None, local_bsz_bounds=None,
-                 elastic_bsz=False):
+                 grad_acc_steps=None, elastic_bsz=False):
```
aurickq, line 56: `grad_acc_steps` should be an internal detail of the SpeedupFunction, i.e. not an argument of `__init__` but an argument of `_goodput` instead.

adaptdl/adaptdl/speedup.py
```diff
@@ -174,13 +182,28 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
+            return true_local_bsz, grad_acc_steps
+        else:
+            return local_bsz, 1
+
     def _goodput(self, nodes, replicas, local_bsz):
+        local_bsz, grad_acc_steps = self._partition_local_bsz(local_bsz)
         log_pred_step_time, _, _ = \
-            _predict_log(self._params, nodes, replicas, local_bsz)
+            _predict_log(self._params, nodes, replicas,
+                         local_bsz, grad_acc_steps)
```
aurickq, line 206: Currently the formula calculates `((accumulation_steps * T_compute) ** gamma + T_network ** gamma) ** (1 / gamma)`, but that assumes the compute-network overlap occurs for every step, which isn't true. Should instead be `(accumulation_steps - 1) * T_compute + (T_compute ** gamma + T_network ** gamma) ** (1 / gamma)`

aurickq (changes requested):

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -63,21 +63,24 @@ class AdaScale(object):
 
     .. _AdaScale: https://proceedings.icml.cc/static/paper_files/icml/2020/4682-Supplemental.pdf
     """  # noqa: E501
-    def __init__(self, optimizer, scale=None, num_replicas=None,
-                 patch_optimizer=False):
+    def __init__(self, optimizer, lr_scheduler=None, scale=None,
```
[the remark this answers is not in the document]
rmfan: Removed

adaptdl/adaptdl/speedup.py
```diff
@@ -256,10 +282,11 @@ def _predict_log(params, nodes, replicas, local_bsz):
...
 
 
-def _predict_compute(params, local_bsz):
+def _predict_compute(params, local_bsz, grad_acc_steps):
     params = Params(*params)
-    # Forward/backward passes should scale linearly with the batch size.
-    return params.alpha_c + params.beta_c * local_bsz
+    # Forward/backward passes should scale linearly with the batch size
+    # and number of accumulation steps.
+    return params.alpha_c + params.beta_c * local_bsz * grad_acc_steps
```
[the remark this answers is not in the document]
rmfan: Fixed

adaptdl/adaptdl/speedup.py
rmfan: This is just a setting for the maximum number of sets. I think it fits in logically better with the other maximum specifications. Maybe this should be renamed tho to indicate better that it's a maximum.

adaptdl/adaptdl/speedup.py
rmfan: Fixed

adaptdl/adaptdl/torch/_metrics.py
```diff
@@ -111,12 +113,14 @@ def _fit_perf_params():
...
     local_bsz = np.array([key[2] for key in state.profile])
+    grad_acc_steps = np.array([key[3] for key in state.profile])
     values = state.profile.values()
     step_time = np.array([val["step_time"] / val["count"] for val in values])
     sync_time = np.array([val["sync_time"] / val["count"] for val in values])
     compute_time = step_time - sync_time
     state.perf_params = adaptdl.speedup.fit(
-            num_nodes, num_replicas, local_bsz, step_time, compute_time)
+        num_nodes, num_replicas, local_bsz, grad_acc_steps,
+        step_time, compute_time)
```
aurickq, line 123: I think there is a problem here. Based on how the metrics are collected, `step_time` contains the average step time across both accumulated and non-accumulated steps. However, the `SpeedupFunction` calculates the sum of step times over a single gradient accumulation cycle.

One suggestion is to collect different `step_time` for accumulated steps and non-accumulated steps. That makes it a bit clearer what's being measured/calculated at all times.

rmfan: Fixed

adaptdl/adaptdl/speedup.py
rmfan: Changed to be a boolean flag on initialization to be whether to allow or disallow gradient accumulation.

adaptdl/adaptdl/torch/data.py
```diff
@@ -205,7 +224,8 @@ def train(self):
         set_batch_size(self.batch_size, self.max_batch_size,
                        self.local_bsz_bounds)
 
-    def autoscale_batch_size(self, max_batch_size, local_bsz_bounds=None):
+    def autoscale_batch_size(self, max_batch_size, local_bsz_bounds=None,
+                             max_acc_grad_steps=1):
```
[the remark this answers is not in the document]
rmfan: Changed to be a boolean flag to allow/disallow gradient accumulation.

adaptdl/adaptdl/torch/_metrics.py
```diff
@@ -45,6 +41,21 @@ def profile_sync_time(sync_time):
 _PREV_REPORT = None
 
 
+def profile_accumulation_commit():
```
aurickq, line 44: Make it a flag `profile_step_commit(accumulate=True)` to reduce logic duplication.

adaptdl/adaptdl/torch/data.py
```diff
@@ -405,11 +451,20 @@ def __iter__(self):
...
                 for idx, batch in enumerate(super().__iter__()):
+                    if idx > iterations:
+                        break
                     with self._elastic.profile():
                         yield batch
                         # Increment by the number of data samples processed
-                        self._elastic.current_index += self.current_batch_size  # noqa: E501
+                        self._elastic.current_index += (
+                            self._elastic._current_local_bsz
+                            * adaptdl.env.num_replicas())
```
aurickq, line 467: I think we should try hard to avoid modifications to `AdaptiveDataLoader`, since that means we are increasing the complexity of implementing custom data loaders. Conceptually, the `AdaptiveDataLoader` can be agnostic to gradient accumulation, and simply produce local batch of the right size, as specified by the `AdaptiveDataLoaderHelper`.

adaptdl/adaptdl/torch/parallel.py
```diff
@@ -64,14 +59,29 @@ def __init__(self, model, optimizer, lr_scheduler=None,
             param.register_hook(functools.partial(self._backward_hook, param))
 
         # Setup for AdaScale, must be after registering backward hooks!
-        self.adascale = AdaScale(optimizer, patch_optimizer=True)
+        self.adascale = AdaScale(optimizer, lr_scheduler, patch_optimizer=True)
```
aurickq, line 62: Passing in `lr_scheduler` is a bug?

adaptdl/adaptdl/speedup.py
```diff
@@ -229,31 +257,36 @@ def fit(nodes, replicas, local_bsz, step_time, step_time_compute):
...
     np = orig_np  # Restore original numpy.
     return Params(*params)
 
 
-def _predict_log(params, nodes, replicas, local_bsz):
+def _predict_log(params, nodes, replicas, local_bsz, grad_acc_steps):
     params = Params(*params)
     step_time_compute = _predict_compute(params, local_bsz)
     step_time_network = _predict_network(params, nodes, replicas)
+    step_time_accumulation = step_time_compute
```
aurickq, line 278: Should this be `step_time_compute * grad_acc_steps`?

adaptdl/adaptdl/speedup.py
```diff
@@ -283,18 +316,21 @@ def _rmse(pred, true):
...
-    log_pred_step_time, pred_step_time_compute, _ = \
-        _predict_log(params, nodes, replicas, local_bsz)
+    log_pred_step_time, pred_step_time_compute, _, pred_grad_acc_time = \
+        _predict_log(params, nodes, replicas, local_bsz, grad_acc_steps)
     # Error of total step time predictions.
     err1 = _rmse(log_pred_step_time, np.log(step_time))
     # Error of compute time predictions.
     err2 = _rmse(np.log(pred_step_time_compute), np.log(step_time_compute))
+    # Error of gradient accumulation
+    err3 = _rmse(np.log(pred_grad_acc_time), np.log(grad_acc_time))
```
aurickq, line 329: We have prior knowledge that the `compute_time` should be the same as (at least similar to) the `step_time` for accumulated steps. So the objective function can be simplified to just two error terms, one for the non-accumulated step time, and one for the accumulated (combined with `compute_time`) step time. That should make this function easier to fit.

adaptdl/adaptdl/speedup.py
```diff
@@ -72,14 +74,19 @@ def __init__(self, params, grad_params=None, init_batch_size=None,
                                     init_batch_size)
         else:
             self._max_batch_size = init_batch_size
+
+        self._true_max_local_bsz = self._max_local_bsz
+        if gradient_accumulation:
+            self._max_local_bsz = self._max_batch_size
```
aurickq, line 80: Confused by this, what's the different between `max_local_bsz` and `true_max_local_bsz` when gradient accumulation is enabled? And why are both required?

aurickq (changes requested):

adaptdl/adaptdl/speedup.py
rmfan: `step_time_accumulation` is normalized by `grad_acc_steps`, to avoid grad(sqrt(x)) evaluated at zero when `grad_acc_steps` is zero.

rmfan: oh, I misread which part this is in

rmfan: the `step_time` computation will need `step_time_accumulation * grad_acc_steps`, but leaving `step_time_accumulation` to be as it is now

adaptdl/adaptdl/torch/parallel.py
rmfan: Yea, leftover from a previous commit. I guess we dont ever need the initial scale of Adascale to be set, given that you can pass in arbitrary objects to it.

adaptdl/adaptdl/speedup.py
rmfan: I need to rename this better tbh. `true_max_local_bsz` is the maximum batch size returned by the dataloader.

rmfan: Removed: changed the logic to a boolean flag

adaptdl/adaptdl/speedup.py
rmfan: Removed

adaptdl/adaptdl/speedup.py
rmfan: Fixed

adaptdl/adaptdl/speedup.py
```diff
@@ -30,7 +30,8 @@
 # is modeled as a constant term plus a retrogression term which increases
 # linearly with the total number of replicas.
 Params = collections.namedtuple("Params", [
-    # T_compute ~ alpha_c + beta_c * local_bsz
+    # T_compute ~ alpha_c + beta_c * local_bsz +
+    #             (alpha_a + beta_a * local_bsz) * grad_acc_steps
```
aurickq, line 34: Comment is outdated?

adaptdl/adaptdl/speedup.py
```diff
@@ -72,14 +73,17 @@ def __init__(self, params, grad_params=None, init_batch_size=None,
                                     init_batch_size)
         else:
             self._max_batch_size = init_batch_size
+
+        self._gradient_accumulation = gradient_accumulation
```
aurickq, line 77: `self._accumulate = accumulate` is short and meaningful enough

adaptdl/adaptdl/speedup_test.py
```diff
@@ -50,27 +53,30 @@ def test_local_bounds():
...
+                                    return_local_bsz=True)
         assert(np.all(bsz >= np.ceil(128 / replicas).astype(int)))
         assert(np.all(np.logical_or(bsz >= (64), speedup == 0.0)))
         assert(np.all(bsz <= (256)))
         assert(np.all(bsz * replicas <= 100 * 128))
         assert(bsz[0] == 128)
+        assert(np.all(steps == 1))
         # multi-node
-        speedup, bsz = fun(replicas, replicas, return_local_bsz=True)
+        speedup, (bsz, steps) = fun(replicas, replicas, return_local_bsz=True)
```
aurickq, line 73: `return_local_bsz` parameter name is no longer accurate since it's returning both bsz and steps now. Maybe `return_config` instead?

adaptdl/adaptdl/speedup.py
```diff
@@ -72,14 +73,17 @@ def __init__(self, params, grad_params=None, init_batch_size=None,
...
             self._params = Params(*params)
         else:
             self._params = None
         if params is not None and init_batch_size is not None:
-            base_step_time, _, _ = _predict_log(self._params,
-                                                np.array([1]), np.array([1]),
-                                                init_batch_size)
+            base_step_time, _, _ = _predict_log(
+                self._params, np.array([1]), np.array([1]),
+                init_batch_size, 1)
```
aurickq, line 86: It's better to let the baseline performance be the best of any configuration using 1 replica. It used to be that the only configuration on 1 replica was to use the initial batch size, because the gradient noise estimation only worked using multiple replicas. However this should not be true anymore with gradient accumulation. So the baseline performance should be set to the best (local_bsz, acc_steps) achievable on 1 replica. See Eqn 15 https://arxiv.org/pdf/2008.12260.pdf.

adaptdl/adaptdl/speedup.py
```diff
@@ -171,18 +176,36 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
-        return ((ret_speedup, ret_local_bsz)
-                if return_local_bsz else ret_speedup)
+            ret_local_bsz = int(ret_local_bsz.item())
+            ret_acc_steps = int(ret_acc_steps.item())
+        if return_local_bsz:
+            return (ret_speedup, (ret_local_bsz, ret_acc_steps))
+        else:
+            return ret_speedup
+
+    def _partition_local_bsz(self, local_bsz):
```
aurickq, line 190: I think some variable renaming is in order. `local_bsz` is exposed through the public API and means the size of the physical batches produced by the dataloader. We should stick with that meaning throughout our code too. Similarly, max_local_bsz should mean the upper-bound of physical batches produced by the dataloader.

We'll need to pick a different name to represent `local_bsz * acc_steps`. Maybe `accum_bsz` or `acc_local_bsz` or something like that?

Currently `local_bsz` inconsistently refers to different things, which makes the code a bit confusing.

adaptdl/adaptdl/speedup.py
```diff
@@ -229,31 +255,35 @@ def fit(nodes, replicas, local_bsz, step_time, step_time_compute):
         params[5] = upper[5] = lower[5]
     if not any(replicas > 2):
         # Fix beta_n and beta_r if no replicas > 2.
-        params[3] = upper[3] = lower[3]
-        params[5] = upper[5] = lower[5]
+        params[3] = upper[3] = lower[5]
+        params[5] = upper[5] = lower[3]
```
aurickq, line 259: Why are the lower bounds swapped here?

aurickq (changes requested):

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -180,11 +205,15 @@ def _reset_avg(self, param_name):
...
-            if self._sum_local_norm is None:
-                self._sum_local_norm = torch.zeros(self._num_params,
-                                                   device=grad.device)
-            self._sum_local_norm[idx] = grad.pow(2).sum()
+        if self._norms is None:
+            self._norms = torch.zeros((self._grad_acc_steps, self._num_params),
+                                      device=grad[0].device)
+        if self._prev_acc_grad is not None:
+            self._norms[self._current_grad_acc_step][idx] = \
+                (grad - self._prev_acc_grad[idx]).pow(2).sum()
```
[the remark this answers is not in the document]
rmfan: Here `grad` should be the pre-accumulated gradients, because this computes the local gradient norms.

adaptdl/adaptdl/torch/_metrics.py
rmfan: Fixed

adaptdl/adaptdl/torch/_metrics.py
```diff
@@ -147,6 +171,7 @@ def __init__(self):
         self.init_batch_size = None
         self.max_batch_size = None
         self.local_bsz_bounds = None
+        self.max_grad_acc_steps = None
```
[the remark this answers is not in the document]
rmfan: Variable removed when refactoring accumulation metrics

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -249,15 +298,25 @@ def step(self, *args, **kwargs):
...
             size = len(param_group["params"])
             grad_sqr = self._state["norm_avg"][offset:offset + size].sum()
             grad_var = self._state["var_avg"][offset:offset + size].sum()
-            gain = (grad_var + grad_sqr) / (grad_var / self._scale + grad_sqr)
+            gain = (grad_var + grad_sqr) / (grad_var / self.scale + grad_sqr)
             param_group["lr"] = gain * param_group["lr"]
             offset += size
+        for group in self._optimizer.param_groups:
+            for p in group["params"]:
+                p.grad /= self._grad_acc_steps
```
aurickq, line 319: Any ideas how the user can apply gradient clipping with auto-accumulation? If they clip before `optimizer.step()` then they will be clipping the pre-averaged gradients. Maybe it's better to divide the gradient in the backward hook instead?

adaptdl/adaptdl/torch/adascale.py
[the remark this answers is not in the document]
rmfan: Fixed

adaptdl/adaptdl/speedup.py
rmfan: Fixed

adaptdl/adaptdl/torch/adascale.py
rmfan: Fixed: moved to backward hook

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -271,5 +329,12 @@ def patch_optimizer(self):
...
             return self.step(*args, **kwargs)
+        old_zero_grad = self._optimizer.zero_grad
 
+        @functools.wraps(self._optimizer.zero_grad)
+        def zero_wrapper(*args, **kwargs):
+            if self._made_step:
+                return old_zero_grad(*args, **kwargs)
+            return None
         self._optimizer.step = wrapper
+        self._optimizer.zero_grad = zero_wrapper
```
aurickq, line 340: I think there's also `model.zero_grad()` which might be used by some training code.

rmfan: Patched model.zero_grad() in `adaptdl/adaptdl/torch/parallel.py`

adaptdl/adaptdl/fit_test.py
```diff
@@ -0,0 +1,63 @@
...
+    result = speedup.fit(nodes, replicas, local_bsz, grad_acc_steps,
+                         step_time, step_time_compute, step_time_accumulation)
+    loss_result = speedup._obj_fn(
+        result, nodes, replicas, local_bsz, grad_acc_steps,
+        step_time, step_time_compute, step_time_accumulation)
+    loss_true = speedup._obj_fn(
+        params, nodes, replicas, local_bsz, grad_acc_steps,
+        step_time, step_time_compute, step_time_accumulation)
+    assert(abs(loss_result - loss_true) < 0.1 * loss_true
+           or loss_result < loss_true)
```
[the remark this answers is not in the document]
rmfan: Added comments

adaptdl/adaptdl/torch/_metrics.py
```diff
@@ -45,25 +41,31 @@ def profile_sync_time(sync_time):
...
-    state.profile[key]["sync_time"] += state.sync_time
-    state.profile[key]["count"] += 1
+    key = (num_nodes, num_replicas,
+           state.local_bsz, state.gradient_accumulation_steps)
+    if accumulation_step:
+        state.profile[key]["accumulation_step_time"] += step_time
+    else:
+        state.profile[key]["step_time"] += step_time
+        state.profile[key]["sync_time"] += state.sync_time
+        state.profile[key]["count"] += 1
```
aurickq, line 57: I think there's a chance of mis-counting if there's a checkpoint-restart during an accumulation step?

sched/adaptdl_sched/allocator.py
```diff
@@ -59,9 +58,7 @@ async def run(self):
             LOG.info("Sleep for 60 seconds")
             await asyncio.sleep(60)
 
-    async def _get_allocations(self):
-        job_list = await self._objs_api.list_namespaced_custom_object(
-            "adaptdl.petuum.com", "v1", "", "adaptdljobs")
+    def _get_allocations(self, job_list):
```
aurickq, line 61: Is this method still needed?

adaptdl/adaptdl/speedup.py
```diff
@@ -91,8 +100,9 @@ def __init__(self, params, grad_params=None, init_batch_size=None,
         self._mem_local_bsz = np.full((self._mem_size, self._mem_size), -1)
         self._mem_speedup[0, 0] = 0.0  # replicas = 0  ==>  speedup = 0
         self._mem_local_bsz[0, 0] = 0
-        self._mem_speedup[1, 1] = 1.0  # replicas = 1  ==>  speedup = 1
-        self._mem_local_bsz[1, 1] = self._init_batch_size
+        if not gradient_accumulation:
+            self._mem_speedup[1, 1] = 1.0  # replicas = 1  ==>  speedup = 1
```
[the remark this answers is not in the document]
rmfan: Oh yea, we set the baseline to be speedup at (1,1) anyways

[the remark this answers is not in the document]
rmfan: By the way we currently return memoized results, I don't think just setting `mem_speedup` will do anything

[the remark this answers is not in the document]
rmfan: Changing to always setting speedup to be (1,1), and the batchsize to be the one computed for the base speedup.

adaptdl/adaptdl/speedup.py
```diff
@@ -135,13 +145,24 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
         elif self._elastic_bsz:
             max_local_bsz = np.floor(self._max_batch_size / replicas)
             min_local_bsz = np.ceil(self._init_batch_size / replicas)
-            if self._max_local_bsz is not None:
+            if (self._max_local_bsz is not None and
+                    not self._gradient_accumulation):
                 max_local_bsz = np.minimum(self._max_local_bsz, max_local_bsz)
             if self._min_local_bsz is not None:
                 min_local_bsz = np.maximum(self._min_local_bsz, min_local_bsz)
+            if self._gradient_accumulation:
```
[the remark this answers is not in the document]
rmfan: Added

adaptdl/adaptdl/speedup.py
```diff
@@ -135,13 +145,24 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
+                original_min_local_bsz = min_local_bsz
+                min_local_bsz = np.where(replicas == 1,
+                                         self._max_local_bsz,
+                                         min_local_bsz)
```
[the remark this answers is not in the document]
rmfan: There's some shenanigans going on here, mostly with how the batchsize candidates for gradient accumulation on a single replica either need to be the initial batch size or force gradient accumulation. I think it will be easier to just rework this to be done by editing `local_bsz` after `local_bsz = np.geomspace(min_local_bsz, max_local_bsz, num=100)` and force `self._partition_local_bsz` to partition when its not the initial batch size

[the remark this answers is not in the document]
rmfan: Reworked this logic

adaptdl/adaptdl/speedup.py
rmfan: shortened the private var. I think the argument should still be the longer version to minimize ambiguity

adaptdl/adaptdl/torch/_metrics.py
rmfan: Fixed

adaptdl/adaptdl/speedup.py
```diff
@@ -217,8 +261,10 @@ def fit(nodes, replicas, local_bsz, step_time, step_time_compute):
     lower = [1e-8, 1e-8] * 3 + [1.0]
     upper = [np.inf, np.inf] * 3 + [10.0]
     if len(np.unique(local_bsz)) == 1:
-        # Fix alpha_c if only observed a single local batch size.
-        params[0] = upper[0] = lower[0]
+        # Fix beta_c if only observed a single local batch size.
+        # This makes the speedup model optimistic with respect to
+        # scaling up the batchsize
+        params[1] = upper[1] = lower[1]
```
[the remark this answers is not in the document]
rmfan: Fixed to have a 50/50 weight between alpha_c, beta_c

adaptdl/adaptdl/speedup.py
```diff
@@ -135,13 +152,22 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
             # Sample a bunch of potential local_bsz values
             local_bsz = np.geomspace(min_local_bsz, max_local_bsz, num=100)
+            local_bsz = np.where(
+                replicas == 1,
+                np.append(
+                    np.broadcast_to(self._init_batch_size,
+                                    (1, max_local_bsz.size)),
+                    np.geomspace(min_local_bsz * 2, max_local_bsz, 99),
+                    axis=0),
+                local_bsz)
```
aurickq, line 170: Still confused about the logic here. When replicas==1, shouldn't we be looking for the best (atomic_bsz, accum_steps) where either (1) atomic_bsz==init_batch_size OR (2) accum_steps>=1? Not sure how the defined range of local_bsz can fully explore those pairs.

adaptdl/adaptdl/speedup.py
```diff
@@ -232,28 +298,32 @@ def fit(nodes, replicas, local_bsz, step_time, step_time_compute):
...
+    step_time_accumulation = step_time_compute
     gamma = params.gamma
     # Return predicted total step time in log-space to avoid numerical issues
     # in autograd and optimization.
-    return (np.log(step_time_compute ** gamma +
-                   step_time_network ** gamma) / gamma,
-            step_time_compute, step_time_network)
+    step_time = np.log(
+        ((step_time_compute) ** gamma + step_time_network ** gamma) **
+        (1 / gamma) + step_time_accumulation * accumulation_steps)
```
aurickq, line 322: I'm not sure about the reason to take the log now that we're exponentiating by (1/gamma) anyways, maybe fine now to just return the prediction in linear space.

adaptdl/adaptdl/torch/data.py
```diff
@@ -405,11 +475,13 @@ def __iter__(self):
...
                 self.batch_sampler.batch_size = self._elastic._sync_local_bsz()
+                iterations = self._elastic.get_iterations(len(self.sampler))
                 for idx, batch in enumerate(super().__iter__()):
+                    if idx > iterations:
+                        break
                     with self._elastic.profile():
                         yield batch
-                        # Increment by the number of data samples processed
-                        self._elastic.current_index += self.current_batch_size  # noqa: E501
+                        self._elastic.increment_index()
```
aurickq, line 484: Is it possible to remove all changes to this class? Otherwise, you'll need to also make sure the BPTTIterator works and test the transformer. And this complexity will extend to all future data loading classes

rmfan: I don't think that is possible, at least not without creating a layer on top of elastic sampler. The primary issue is that the length of the dataset really needs to be zero mod accumulation steps. I think BPTTIterator will need to be changed for it to work with gradient accumulation, but it should be a one line change.

adaptdl/adaptdl/speedup.py
```diff
@@ -72,15 +76,24 @@ def __init__(self, params, grad_params=None, init_batch_size=None,
...
-            base_step_time = base_step_time.item()
+            if self._accumulation:
+                max_steps = int(self._max_batch_size / init_batch_size)
+            else:
+                max_steps = 0
+            step_times = _predict_log(
+                self._params, np.ones((max_steps + 1,)),
+                np.ones((max_steps + 1,)), init_batch_size,
+                np.asarray(range(max_steps + 1)))[0]
+            base_step_time = np.max(step_times).item()
```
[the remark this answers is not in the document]
rmfan: Fixed: changed it to use goodput (which partitions properly automatically). Now it should choose the best from [init_batch_size, 2*init_batch_size, ..., max_batch_size]. Maybe there could be higher resolution there tho,.

adaptdl/adaptdl/speedup.py
rmfan: Changed

adaptdl/adaptdl/speedup.py
```diff
@@ -171,31 +202,66 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
+    def _partition_local_bsz(self, local_bsz, replicas):
+        if (self._max_local_bsz is not None and (
+                np.any(self._max_local_bsz < local_bsz)
+                or np.any(replicas == 1))):
+            accumulation_steps = np.maximum(
+                np.ceil(local_bsz / self._max_local_bsz),
+                1)
+            atomic_bsz = np.minimum(
+                self._max_local_bsz, np.ceil(local_bsz / accumulation_steps))
+            for i, replica in enumerate(replicas):
```
aurickq, line 226: Can this loop be vectorized?

adaptdl/adaptdl/speedup.py
```diff
@@ -171,31 +194,78 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
+                np.any(self._max_local_bsz < local_bsz)
+                or np.any(replicas == 1))):
+            accumulation_steps = np.maximum(
+                np.ceil(local_bsz / self._max_local_bsz),
+                1)
+            atomic_bsz = np.minimum(
+                self._max_local_bsz, np.ceil(local_bsz / accumulation_steps))
+
+            # Need to prevent a single replica from scaling up the
+            # batch size without using gradient accumulation
```
aurickq, line 220: Feels like the following code can be greatly simplified using index-assign operations instead of functional style. Should be fine since this function is not evaluated using autograd I think.

adaptdl/adaptdl/speedup.py
```diff
@@ -171,31 +194,78 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
+            single_replica_predicate = replicas == 1
+            single_step_predicate = np.logical_or(
+                local_bsz == self._init_batch_size,
+                local_bsz <= 2 * self._min_local_bsz)
```
aurickq, line 224: Is the predicate just `(replicas == 1) & (accumulation_steps <= 1)`?

adaptdl/adaptdl/speedup.py
```diff
@@ -171,31 +194,78 @@ def __call__(self, nodes, replicas, return_local_bsz=False):
...
-        return ((ret_speedup, ret_local_bsz)
-                if return_local_bsz else ret_speedup)
+            ret_atomic_bsz = int(ret_atomic_bsz.item())
+            ret_accumulation_steps = int(ret_accumulation_steps.item())
+        if return_config:
+            return (ret_speedup, (ret_atomic_bsz, ret_accumulation_steps))
+        else:
+            return ret_speedup
+
+    def _partition_local_bsz(self, local_bsz, replicas):
```
aurickq, line 209: I think this method needs an explanation comment.

adaptdl/adaptdl/speedup.py
```diff
@@ -72,17 +76,31 @@ def __init__(self, params, grad_params=None, init_batch_size=None,
...
-            base_step_time, _, _ = _predict_log(self._params,
-                                                np.array([1]), np.array([1]),
-                                                init_batch_size)
-            base_step_time = base_step_time.item()
-            self._base_goodput = 1.0 / np.exp(base_step_time)
+            if self._accumulation:
+                max_steps = int(self._max_batch_size / init_batch_size)
+                candidates = np.arange(init_batch_size,
+                                       self._max_batch_size + 1,
+                                       init_batch_size)
```
aurickq, line 91: Can just use `geomspace` like in `__call__` to increase the resolution?

adaptdl/adaptdl/torch/data.py
```diff
@@ -195,6 +196,58 @@ def local_bsz_bounds(self):
...
+    @property
+    def is_accumulation_step(self):
+        """
+        True iff the current step is a gradient accumulation step:
+        I.e. if the model parameters aren't updated
+        """
+        return self._is_accumulation_step
+
+    @is_accumulation_step.setter
+    def is_accumulation_step(self, value: bool):
```
aurickq, line 225: Don't think you need a setter for this since it's only being modified from the same class?

adaptdl/adaptdl/speedup.py
rmfan: I think that would result in slightly different behavior. With `(replicas == 1) & (accumulation_steps <= 1)`, then that would result in batch sizes in between `self._init_batch_size` and `max_local_bsz` being divided up into `(self._init_batch_size, 1)`, instead of `(batch_size/2, 2)`. I think the second is preferable.

adaptdl/adaptdl/speedup.py
rmfan: Added

adaptdl/adaptdl/speedup.py
rmfan: Switched to using `switch` instead of where for increased readability.

adaptdl/adaptdl/torch/data.py
rmfan: Its mostly modified in `parallel.py`. The time its modified in `data.py` is so that the value is correct in the very first step after a `_sync_local_bsz` before it can be updated by `forward` in `parallel.py`.

rmfan: `adascale.py` is treated as the truth for whether the program is in an accumulation step or not, and given that the dataloader needs that in a context manager, its difficult to pass it through without setting a variable.

rmfan:
Speedup model fit data:
speedup

aurickq (approved):

PR #29, opened 2020-09-18
Push docker images through local mitmproxy to remote insecure registry

### LINKED ISSUE #21: Remove sudo requirement from adaptdl submit
Currently, the adaptdl submit command requires sudo to modify the /etc/hosts file to assign a consistent name to the insecure docker registry included with the AdaptDL scheduler. This consistent name is added as an insecure registry to the Docker daemon.json and as a tag for built images. This is not ideal since, for security reasons, we don't want the CLI to use sudo on client machines.

One potential solution is to (for the duration of the submit command) proxy/port-forward the insecure registry to `localhost`, which should be automatically trusted by the local Docker client, thus no longer necessary to modify daemon.json.

aurickq:
Currently, when the built-in insecure registry is used, `adaptdl submit` will push the image directly to the registry through a NodePort service. Since the remote node has a dynamic address and is an insecure Docker registry, `adaptdl submit` needs to add the remote address to the local /etc/hosts file, /etc/docker/daemon.json file, and restart the docker daemon. All three of these actions require admin privileges.

With this PR, `adaptdl submit` will instead launch a proxy on localhost, which is a consistent address automatically trusted by the Docker client. The image is then pushed through the proxy to the remote insecure registry.

resolves #21 

Also adds a simple command to proxy to TensorBoard instances: `adaptdl tensorboard proxy <my-tensorboard>`.


aurickq:
Tested on EKS

adaptdl_cli/adaptdl_cli/proxy.py
```diff
@@ -0,0 +1,92 @@
...
+        namespace (str): namespace of the target Kubernetes service.
+        service (str): name of the target Kubernetes service, in the form
+            [https:]service_name[:port_name].
+        listen_host (str): local address to bind the proxy.
+        listen_port (int): local port to bind the proxy. If 0, selects an
+            arbitrary free port.
+        verbose (bool): if True, prints extra logs from the background proxy.
+    """
+    listen_port = listen_port if listen_port else pick_unused_port()
+    queue = Queue()
```
odp, line 40: A barrier is preferred here instead of a queue.

odp:
How's the error condition handled when there are no pods running behind the registry-service? This can happen on EKS when the nodes are taken down (because of various reasons like spot preemption or company policy) but control plane is up (so you see the service)

aurickq:
> How's the error condition handled when there are no pods running behind the registry-service? This can happen on EKS when the nodes are taken down (because of various reasons like spot preemption or company policy) but control plane is up (so you see the service)

The proxy won't be able to forward any requests, so I think it will just crash.

odp:
> > How's the error condition handled when there are no pods running behind the registry-service? This can happen on EKS when the nodes are taken down (because of various reasons like spot preemption or company policy) but control plane is up (so you see the service)
> 
> The proxy won't be able to forward any requests, so I think it will just crash.

Want to make sure that we are not left with orphan processes and somehow user's would know what the problem is instead of seeing some random stack trace.

aurickq:
Also tested submitting to EKS from both Ubuntu and MacOS clients. Had to make a fix for MacOS, since Docker for Mac runs inside a VM and `docker push` could not access the proxy on the host. So for Mac we still need to use a different image name and manually add it as an insecure registry.

adaptdl_cli/bin/adaptdl
```diff
@@ -17,86 +17,115 @@
...
+            with open(daemon_path, "r") as f:
+                daemon = json.load(f)
+        except FileNotFoundError:
+            daemon = {}
+        if registry not in daemon.setdefault("insecure-registries", []):
+            daemon["insecure-registries"].append(registry)
+            docker_dir.mkdir(exist_ok=True)
+            with open(daemon_path, "w") as f:
+                json.dump(daemon, f, indent=2)
+            raise SystemExit(f"Insecure registry {registry} written to"
```
odp, line 81: If the user forgets to restart docker or doesn't do it. Would this run silently on the his next try?

adaptdl_cli/bin/adaptdl
```diff
@@ -320,6 +348,9 @@ if __name__ == "__main__":
     parser_submit.add_argument(
         "--checkpoint-storage-class", type=str,
         help="storage class of new temporary pvc to create", default=None)
+    parser_submit.add_argument(
+        "--proxy-port", type=int, default=59283,  # Some uncommon port.
```
odp, line 352: Not sure why this an argument and not just any available port (using say portpicker)?

aurickq:
> > > > How's the error condition handled when there are no pods running behind the registry-service? This can happen on EKS when the nodes are taken down (because of various reasons like spot preemption or company policy) but control plane is up (so you see the service)
> > > 
> > > 
> > > The proxy won't be able to forward any requests, so I think it will just crash.
> > 
> > 
> > Want to make sure that we are not left with orphan processes and somehow user's would know what the problem is instead of seeing some random stack trace.
> 
> Ok so what happens in my test is, no crashes, but Docker will not be able to push, resulting in output like the following:
> 
> ```
> Sending build context to Docker daemon  77.31kB
> Step 1/4 : FROM pytorch/pytorch:1.4-cuda10.1-cudnn7-runtime
>  ---> 37b81722dadc
> Step 2/4 : WORKDIR /root
>  ---> Using cache
>  ---> fac22f772151
> Step 3/4 : RUN pip install adaptdl torch torchvision tensorboard
>  ---> Using cache
>  ---> b96faab58673
> Step 4/4 : COPY . .
>  ---> Using cache
>  ---> dae17decfa13
> Successfully built dae17decfa13
> Successfully tagged host.docker.internal:59283/adaptdl-submit:latest
> The push refers to repository [host.docker.internal:59283/adaptdl-submit]
> 166c23429ae8: Retrying in 9 seconds
> 6a41d54cf8d9: Retrying in 9 seconds
> bd03b36e2f09: Retrying in 9 seconds
> 3fc0d5aee71e: Retrying in 9 seconds
> d39a607867ff: Retrying in 9 seconds
> d5bebe868b18: Waiting
> 6670ca71777c: Waiting
> 918efb8f161b: Waiting
> 27dd43ea46a8: Waiting
> 9f3bfcc4a1a8: Waiting
> 2dc9f76fb25b: Waiting
> ```
> 
> After a few retries, Docker will fail with an error and the command exits, like the following:
> 
> ```
> The push refers to repository [host.docker.internal:59283/adaptdl-submit]
> 166c23429ae8: Retrying in 1 second
> 6a41d54cf8d9: Retrying in 1 second
> bd03b36e2f09: Retrying in 1 second
> 3fc0d5aee71e: Retrying in 1 second
> d39a607867ff: Retrying in 1 second
> d5bebe868b18: Waiting
> 6670ca71777c: Waiting
> 918efb8f161b: Waiting
> 27dd43ea46a8: Waiting
> 9f3bfcc4a1a8: Waiting
> 2dc9f76fb25b: Waiting
> received unexpected HTTP status: 503 Service Unavailable
> Traceback (most recent call last):
>   File "bin/adaptdl", line 394, in <module>
>     args.handler(args, remaining)
>   File "bin/adaptdl", line 116, in submit
>     repodigest = _build_push(args.project, args.dockerfile, args.proxy_port)
>   File "bin/adaptdl", line 104, in _build_push
>     subprocess.check_call(["docker", "push", image])
>   File "/Users/aurickq/opt/miniconda3/envs/adaptdl/lib/python3.8/subprocess.py", line 364, in check_call
>     raise CalledProcessError(retcode, cmd)
> subprocess.CalledProcessError: Command '['docker', 'push', 'host.docker.internal:59283/adaptdl-submit']' returned non-zero exit status 1.
> ```
> 
> I think this behavior should be fine for now?

Now looks like:
```
The push refers to repository [host.docker.internal:59283/adaptdl-submit]
166c23429ae8: Retrying in 1 second
6a41d54cf8d9: Retrying in 1 second
bd03b36e2f09: Retrying in 1 second
3fc0d5aee71e: Retrying in 1 second
d39a607867ff: Retrying in 1 second
d5bebe868b18: Waiting
6670ca71777c: Waiting
918efb8f161b: Waiting
27dd43ea46a8: Waiting
9f3bfcc4a1a8: Waiting
2dc9f76fb25b: Waiting
received unexpected HTTP status: 503 Service Unavailable
Error: could not push image host.docker.internal:59283/adaptdl-submit
```

adaptdl_cli/bin/adaptdl
aurickq: It will fail at the docker push step

adaptdl_cli/bin/adaptdl
aurickq: If we pick a different port for each `adaptdl submit` then it will leave many temporary image tags hanging around in `docker images`. Also in the MacOS case it will require modifying daemon.json and restarting docker each time `adaptdl submit` is run.

adaptdl_cli/bin/adaptdl
odp: yes but from the next try onwards users won't see this message and would never know what they missed (?)

aurickq: The next time, the following error is printed:
```
The push refers to repository [host.docker.internal:55555/adaptdl-submit]
Get https://host.docker.internal:55555/v2/: x509: certificate signed by unknown authority
Error: could not push image host.docker.internal:55555/adaptdl-submit
```
Googling for "docker x509: certificate signed by unknown authority" returns as the first result a stackoverflow answer suggesting to restart docker. I think that's fine unless there's a better way for us to detect that a Docker restart is needed.

adaptdl_cli/bin/adaptdl
odp: In case the default port is not available on localhost, the user has to choose an available port. Could they use the same `proxy-port` for multiple successive `adaptdl submit` runs without getting port bind errors (from the proxy)?

adaptdl_cli/bin/adaptdl
odp: I think we should document this as part of some FAQs

adaptdl_cli/bin/adaptdl
aurickq: Yes

adaptdl_cli/bin/adaptdl
aurickq: Added

odp (approved):
LGTM

odp (approved):

PR #35, opened 2020-09-24
Pollux policy cluster utility bugfix

aurickq:
- Fix cluster utility calculation to only consider nodes which have at least one pod scheduled on them.
- Fix/simplify Makefile to use cli proxy to push scheduler image.


sched/adaptdl_sched/policy/pollux.py
```diff
@@ -241,9 +241,8 @@ def get_cluster_utilities(self, states):
         """
         num_replicas = np.sum(states, axis=2)
         speedups = self._get_job_speedups(states)
-        cluster_sizes = self._get_cluster_sizes(states)
         # mask (pop_size x num_nodes): indicates which nodes are active.
-        mask = np.arange(len(self._nodes)) < np.expand_dims(cluster_sizes, 1)
+        mask = np.sum(states, axis=1) > 0
```
odp, line 245: Can you add few test cases to Pollux that cover the issue that was reported? May be add few other combinations?

Makefile
```diff
@@ -1,64 +1,41 @@
-ifeq "$(and $(ADAPTDL_DEV_REPO),$(ADAPTDL_DEV_REPO_CREDS))" ""
-	IMAGE_REPOSITORY := $(shell python3 -c 'from adaptdl_cli.adaptdl_cli.registry import ADAPTDL_REGISTRY_URL; print(ADAPTDL_REGISTRY_URL)')/dev/adaptdl-dev
-	REPO_CREDS := $(shell python3 -c 'from adaptdl_cli.adaptdl_cli.registry import ADAPTDL_REGISTRY_CREDS; print(ADAPTDL_REGISTRY_CREDS)')
-else
-	IMAGE_REPOSITORY := $(ADAPTDL_DEV_REPO)
```
odp, line 5: How does one use external secure repo for developer workflow?. It might be desirable for concerns like speed and persistence.

odp (changes requested):

Makefile
aurickq: Done, `ADAPTDL_DEV_REPO` and `ADAPTDL_DEV_REPO_CREDS` same as before.

sched/adaptdl_sched/policy/pollux.py
aurickq: Added a unit test. Confirmed it fails with the previous code and works with the updated code.

Makefile
```diff
@@ -1,64 +1,45 @@
...
 		--set image.digest=$(IMAGE_DIGEST) \
-		--set image.secrets.name=$(REPO_CREDS) \
-		--set docker-registry.enabled=true \
 		--values .values.yaml
 
-delete: .devtag
+delete:
 	helm delete $(RELEASE_NAME)
-
-pip-install-reqs:
```
odp, line 47: The dependencies of proxy service like portpicker and mitmproxy should be checked before calling the proxy. Currently they will be installed only with `adaptdl-cli` package which is not really a dependency of `make deploy`.

odp (changes requested):

Makefile
aurickq: Ok now `make deploy` without the necessary requirements will error out with:
```
(adaptdl) $ make deploy
The 'mitmproxy>=5.2' distribution was not found and is required by the application
To install requirements, run the command:
/miniconda3/envs/adaptdl/bin/python3 -m pip install -r adaptdl_cli/requirements.txt
Makefile:27: recipe for target 'check-requirements' failed
make: *** [check-requirements] Error 1
```

odp (dismissed):
LGTM

odp (approved):

PR #37, opened 2020-09-26
Separate validation logic out into a ValidatingAdmissionWebhook

### LINKED ISSUE #25: Validate AdaptDLJobs using an AdmissionWebhook
Currently, AdaptDLJob CRD validation occurs in the controller, after the custom resource is already created. This has several drawbacks:

- It further complicates the already-complicated controller code.
- Users creating invalid AdaptDLJob only gets feedback after it is successfully created and transitions to the "Failed" phase.
- It's difficult to prevent the job spec from being modified after it's created. Currently the controller will let modifications occur, but fail the job if a modification is detected.

Instead, we can separate the validation logic into a ValidatingAdmissionWebhook, which validates newly-created AdaptDLJobs and initializes their status sub-resources. Invalid jobs will immediately result in an error response from the job creation step, and modifications to job specs can be prevented without failing the job entirely.

ValidatingAdmissionWebhook requires TLS. We might be able to auto-generate a certificate at the helm install step using genSelfSignedCert sprig function: http://masterminds.github.io/sprig/crypto.html.

References:

- https://medium.com/nuvo-group-tech/move-your-certs-to-helm-4f5f61338aca
- https://gitlab.com/ktarplee/gpu-admission-webhook/-/blob/master/charts/gpu-admission-webhook/templates/webhook.yaml#L12

aurickq:
Resolves #25 

With this PR, creating an AdaptDLJob with an invalid pod template will fail at creation time rather than marked as a job failure:
```
Failure (Invalid): admission webhook "adaptdl-validator.default.svc.cluster.local" denied the request: PodTemplate "spec.template" is invalid: template.spec.containers[0].name: Required value
```

The webhook will also prevent job specs from being modified after the job is already created.


sched/adaptdl_sched/validator.py
```diff
@@ -0,0 +1,132 @@
...
+            "template": job["spec"]["template"],
+        }
+        try:
+            await self._core_api.create_namespaced_pod_template(
+                namespace, template, dry_run="All")
+        except kubernetes.client.rest.ApiException as exc:
+            return {
+               "allowed": False,
+               "status": {
+                   "code": 400,
```
odp, line 83: Can you use [symbolic](https://docs.python.org/3/library/http.html) status codes everywhere?

odp (changes requested):

sched/adaptdl_sched/validator.py
aurickq: Ok done

odp (approved):
LGTM

PR #48, opened 2020-10-16
Issue 27

YLJALDC:


adaptdl/adaptdl/torch/data.py
```diff
@@ -239,10 +249,26 @@ def _sync_local_bsz(self):
...
+
+            # if first time called, initiliza the batch size field
+            if(self.init_time):
+                self.cur_local_bsz = local_bsz
+                self.init_time = False
+
+            # if the speedup is significant, we use it, 
+            # otherwise, keep the old one.
+
+            if(speedup > self.speedup_threshold):
```
aurickq, line 264: Doesn't this just accept any batch size with a speedup of 1.05 or greater? What we want is to only change the batch size if the _increase_ in speedup over the current speedup surpasses some threshold.

aurickq (changes requested):

adaptdl/adaptdl/torch/data.py
```diff
@@ -147,6 +147,16 @@ def __init__(self, batch_size=1):
         self.batch_size = batch_size
         self.future_exit = None
 
+        # True before the first time sync_local_bsz.
+        self.init_time = True
+
+        # initialize a record of current local bs, if the sppedup 
+        # is not significant then we will keep the previous batch size
+        self.cur_local_bsz = 0
```
aurickq, line 155: Don't think you need to introduce these two new fields. There's the existing `current_local_bsz` which you can just initialize to `None` here instead.

PR #49, opened 2020-10-17
Issue 27

YLJALDC:


adaptdl/adaptdl/torch/data.py
```diff
@@ -239,10 +245,40 @@ def _sync_local_bsz(self):
...
             speedup_fn = get_speedup_fn()
-            _, local_bsz = speedup_fn(adaptdl.env.num_nodes(),
-                                      adaptdl.env.num_replicas(),
-                                      return_local_bsz=True)
-            self.current_local_bsz = adaptdl.collective.broadcast(local_bsz)
+
+            # check whether predict param is up
+            is_init = speedup_fn._params is None
+            # if first time called, initiliza the batch size field
+            if is_init:
```
aurickq, line 252: I think you can determine this case simply with `self.current_local_bsz is None`? Accessing private field `_params` is not a good practice.

aurickq (changes requested):

Issue #9, opened 2020-09-01
Export useful metrics to TensorBoard


aurickq:
It would be nice to include a helper function that bulk-exports all the relevant AdaptDL metrics to TensorBoard, so applications won't need to manually select and expose individual AdaptDL metrics. Some examples:

- Number of replicas
- Current goodput and throughput
- Estimated gradient noise scale
- Gain from AdaScale

Having these can improve the understandability of what AdaptDL is doing.

PR #56, opened 2020-10-21
Add methods to output tensorboard metrics

aurickq:
Resolves https://github.com/petuum/adaptdl/issues/9

Add a few helper methods to output relevant tensorboard metrics.

### DIFF
--- a/adaptdl/adaptdl/torch/data.py
+++ b/adaptdl/adaptdl/torch/data.py
@@ -365,6 +365,8 @@ class AdaptiveDataLoaderHelper(object):
             global_step (int): Global step value to record.
             tag_prefix (str): Prefix added to each metric's tag.
         """
+        if tag_prefix and not tag_prefix.endswith("/"):
+            tag_prefix += "/"
         writer.add_scalar(tag_prefix + "Total_Batch_Size",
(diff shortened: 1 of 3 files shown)


adaptdl/adaptdl/torch/data.py
```diff
@@ -355,6 +355,23 @@ def skipdone(self):
...
+        """
+        Output some useful metrics to TensorBoard.
+
+        Arguments:
+            writer (torch.utils.tensorboard.SummaryWriter): ``SummaryWriter``
+                object to output metrics to.
+            global_step (int): Global step value to record.
+            tag_prefix (str): Prefix added to each metric's tag.
+        """
+        writer.add_scalar(tag_prefix + "Total_Batch_Size",
```
rmfan, line 368: I think this should be joined together with a `/`, otherwise tag_prefix will need to end with a `/`, which doesnt seem too intuitive to me.

rmfan: Alternatively, adding a `/` if the tag doesnt end in a `/` already

rmfan (changes requested):

rmfan (approved):

PR #58, opened 2020-10-22
add speedup ratio comparison

YLJALDC:
Added speedup ratio comparison. If suggest_speedup / current_speedup < 1.05, the batch size will not change. 
Orange is original implementation. Pink is the new feature.
Capture
Capture
Capture

### DIFF
--- a/adaptdl/adaptdl/torch/data.py
+++ b/adaptdl/adaptdl/torch/data.py
@@ -147,7 +147,7 @@ class AdaptiveDataLoaderHelper(object):
         adaptdl.checkpoint.load_state(self._state)
         self.batch_size = batch_size
         self.future_exit = None
-        self._gradient_accumulation = False
+        self.speedup_threshold = 1.05
 
     @property
     def current_index(self):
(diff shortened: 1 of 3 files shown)


adaptdl/adaptdl/torch/data.py
```diff
@@ -147,7 +147,7 @@ def __init__(self, batch_size=1):
         adaptdl.checkpoint.load_state(self._state)
         self.batch_size = batch_size
         self.future_exit = None
-        self._gradient_accumulation = False
```
aurickq, line 150: Don't think this line should be deleted

adaptdl/adaptdl/torch/data.py
```diff
@@ -276,12 +276,40 @@ def _sync_local_bsz(self):
...
+                speedup_to_cur = suggest_speedup / current_speedup
+                if speedup_to_cur > self.speedup_threshold:
+                    (self._current_local_bsz, self._accumulation_steps) = \
+                        adaptdl.collective.broadcast((local_bsz,
+                                                      accumulation_steps))
+                else:
+                    (self._current_local_bsz, self._accumulation_steps) = \
+                        adaptdl.collective.broadcast((
+                            self._current_local_bsz,
+                            self._accumulation_steps))
```
aurickq, line 311: It might be better to run
```
self._current_local_bsz, self._accumulation_steps) = \
    adaptdl.collective.broadcast((
        self._current_local_bsz,
        self._accumulation_steps))
```
just after all the if-else branches, and simply assign `self._current_local_bsz` and `self._accumulation_steps` in the if-else branches. That way it's a little less verbose and makes it clear that the broadcast is always called exactly once on each replica.

adaptdl/adaptdl/torch/data.py
```diff
@@ -276,12 +276,40 @@ def _sync_local_bsz(self):
...
             speedup_fn = get_speedup_fn()
-            _, (local_bsz, accumulation_steps) = speedup_fn(
-                adaptdl.env.num_nodes(),
-                adaptdl.env.num_replicas(),
-                return_config=True)
-            (self._current_local_bsz, self._accumulation_steps) = \
-                adaptdl.collective.broadcast((local_bsz, accumulation_steps))
+
+            # check if batch size  is up
+            is_init = self._current_local_bsz is None
```
aurickq, line 282: No need to assign to a variable if it's just used once in the next line

adaptdl/adaptdl/torch/data.py
YLJALDC: changed

aurickq (approved):

PR #61, opened 2020-10-26
Support client-cert auth for mitmproxy

aurickq:


cli/adaptdl_cli/proxy.py
```diff
@@ -76,10 +77,22 @@ def _run_proxy(event, namespace, service, listen_host, listen_port, verbose):
     master = DumpMaster(options, with_termlog=verbose, with_dumper=verbose)
     options.keep_host_header = True
     options.ssl_insecure = True
```
odp, line 79: Should this be `False` in case we have a cert and a key below?

odp (changes requested):

cli/adaptdl_cli/proxy.py
aurickq: I don't think so... This field is for the mitmproxy to verify the Kubernetes ApiServer's certificate. The change in this PR is to use the client's certificate to authenticate if available.

odp (approved):

PR #64, opened 2020-10-30
Issue 19

YLJALDC:


sched/adaptdl_sched/policy/pollux.py
```diff
@@ -40,6 +40,32 @@ def __init__(self):
...
+                   for key, val in job_resources.items())
+            # we are done if addding this node gives enough replicas    
+            if (len(node_list) + replica_this) >= min_replicas:
+                replica_need = min_replicas - len(node_list) 
+                node_list.extend([node_name] * replica_need)
+                LOG.info("find %s for job %s with minimum replicas: %s",
+                         node_list, job_key, min_replicas)    
+                return node_list
+            # else we replicate as many as we can on this node
+            else:
```
aurickq, line 62: Hmm, this can violate the Pollux restriction that no two distributed jobs (i.e. each having replicas on multiple nodes) may share the same node (having a replica on the same node). Maybe it's best to just search for resources on a single node instead of allowing the new job to span multiple nodes.

sched/adaptdl_sched/allocator.py
```diff
@@ -37,10 +38,58 @@ class AdaptDLAllocator(object):
...
+                                 "", "adaptdljobs")
         self._cluster_expander = expander
         self._policy = PolluxPolicy()
 
     async def run(self):
+        # two functionality: (1) watch for new job and start if possible.
+        # (2) periodically optimize existing jobs
+        await asyncio.gather(
+            self._new_job_allocate(),
+            self._periodic_optimize()
```
aurickq, line 51: Need to lock the inner loops so they don't happen concurrently.

aurickq (changes requested):

sched/adaptdl_sched/policy/pollux.py
```diff
@@ -40,6 +40,32 @@ def __init__(self):
         self._min_util = 0.35
         self._max_util = 0.65
 
+    # a simple allocate policy: allocate the new job 
+    # to the first available node.
```
aurickq, line 44: Unit test this function

PR #66, opened 2020-10-30
added job trigger with synchronization

YLJALDC:


sched/adaptdl_sched/allocator.py
```diff
@@ -38,11 +38,90 @@ class AdaptDLAllocator(object):
...
+                                            job_info, node_info_list)
+                        # allocate the job on the suggest_node
+                        if new_allocation:
+                            patch = {"status": {"allocation": new_allocation}}
+                            LOG.info("Patch AdaptdlJob %s/%s: %s ",
+                                     namespace, name, patch)
+                            await patch_job_status(self._objs_api, namespace,
+                                                   name, patch)
+                        # release lock
+                        self.lock.release()
```
aurickq, line 119: Use `async with` rather than `acquire` and `release`

aurickq: Also suggest splitting out the inner contents of the loop into a different method.

sched/adaptdl_sched/allocator.py
```diff
@@ -38,11 +38,90 @@ class AdaptDLAllocator(object):
...
+                    if event["type"] == "ADDED":
+                        await self.lock.acquire()
+                        # re-read the job , compared with the previous readjob
+                        job_old = event["object"]
+                        namespace_old = job_old["metadata"]["namespace"]
+                        name_old = job_old["metadata"]["name"]
+
+                        job_list = await \
+                            self._objs_api.list_namespaced_custom_object(
+                                "adaptdl.petuum.com", "v1", "", "adaptdljobs")
```
aurickq, line 72: Just read the single job object rather than list every job.

sched/adaptdl_sched/allocator.py
```diff
@@ -38,11 +38,90 @@ class AdaptDLAllocator(object):
...
+                                break
+                        else:
+                            # This job is don
+                            self.lock.release()
+                            continue
+
+                        # some other coroutine has handled this
+                        LOG.info(job.get("status", {}).get("allocation", []))
+                        if len(job.get("status", {})
+                                .get("allocation", [])) != 0:
```
aurickq, line 91: Just `not job.get("status", {}).get("allocations")` is enough.

Also I think we need to add a `job.get("status", {}).get("group") is None` because the current condition doesn't catch the case where a job is allocated some resources, de-allocated all resources, and allocated again.

sched/adaptdl_sched/allocator.py
```diff
@@ -102,62 +184,66 @@ async def _find_nodes(self):
         node_template = NodeInfo(max_resources, True)
         return node_infos, node_template
 
+    def _get_job_info(self, job, job_infos, allocations=None):
+        if allocations and job.get("status", {}).get("phase") \
+                not in ["Pending", "Running", "Starting", "Stopping"]:
+            LOG.info("unkown status:", job, job.get("status", {}).get("phase"))
+            return
```
aurickq, line 191: Maybe pull this logic out, and only call `_get_job_info` if it's the right status.

sched/adaptdl_sched/allocator.py
```diff
@@ -102,62 +184,66 @@ async def _find_nodes(self):
...
+                hints.get("gradientAccumulation", False))
+        else:
+            speedup_fn = lambda n, r: r  # noqa: E731
+        creation_ts = dateutil.parser.isoparse(
+                job["metadata"]["creationTimestamp"])
+        namespace = job["metadata"]["namespace"]
+        name = job["metadata"]["name"]
+        job_infos[(namespace, name)] = JobInfo(
+                resources, speedup_fn, creation_ts, min_replicas,
+                max_replicas, preemptible)
```
aurickq, line 238: Return the requested info rather than modifying an argument.

sched/adaptdl_sched/allocator.py
```diff
@@ -102,62 +184,66 @@ async def _find_nodes(self):
         node_template = NodeInfo(max_resources, True)
         return node_infos, node_template
 
+    def _get_job_info(self, job, job_infos, allocations=None):
+        if allocations and job.get("status", {}).get("phase") \
+                not in ["Pending", "Running", "Starting", "Stopping"]:
```
aurickq, line 189: This condition doesn't seem to be the same as before. Why is it changed?

sched/adaptdl_sched/allocator.py
```diff
@@ -38,11 +38,90 @@ class AdaptDLAllocator(object):
...
+                                     namespace, name, patch)
+                            await patch_job_status(self._objs_api, namespace,
+                                                   name, patch)
+                        # release lock
+                        self.lock.release()
+
+    async def _periodic_optimize(self):
         while True:
+            # try to gain lock
+            await self.lock.acquire()
```
aurickq, line 124: Use `async with` and maybe pull out the inner contents into a separate method

aurickq (changes requested):

sched/adaptdl_sched/allocator.py
```diff
@@ -102,62 +184,66 @@ async def _find_nodes(self):
...
+    def _get_job_info(self, job, job_infos, allocations=None):
+        if allocations and job.get("status", {}).get("phase") \
+                not in ["Pending", "Running", "Starting", "Stopping"]:
+            LOG.info("unkown status:", job, job.get("status", {}).get("phase"))
+            return
+        if allocations and "allocation" in job.get("status", {}):
+            namespace = job["metadata"]["namespace"]
+            name = job["metadata"]["name"]
+            allocations[namespace, name] = \
+                list(job["status"]["allocation"])
```
aurickq, line 196: Return requested info rather than modifying the arguments

sched/adaptdl_sched/allocator.py
YLJALDC: allocations is None means that I am only parsing one job (called from allocate job), so it should get rid of this phase condition because that phase is async and shouldnt matter.

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,109 @@ class AdaptDLAllocator(object):
...
+    async def _new_job_allocate(self):
+        async with kubernetes.watch.Watch() as watch:
+            while True:
+                async for event in watch.stream(
+                        self._objs_api.list_namespaced_custom_object,
+                        *self._custom_resource, timeout_seconds=60):
+                    self._new_job_helper(event)
+                   
+                   
+    async def _new_job_helper(self):
```
aurickq, line 65: To keep these method names more intuitive, maybe `_allocate_one`, `_allocate_one_loop`, `_optimize_all`, and `_optimize_all_loop`?

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,109 @@ class AdaptDLAllocator(object):
...
+        # if there is an arriving job
+        if event["type"] == "ADDED":
+            async with self.lock:
```
aurickq, line 68: Lock and if statement can go outside of this method, so this method can be more unit-testable in the future

aurickq (changes requested):

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,93 @@ class AdaptDLAllocator(object):
     def __init__(self, expander):
         self._core_api = kubernetes.client.CoreV1Api()
         self._objs_api = kubernetes.client.CustomObjectsApi()
+        self._custom_resource = ("adaptdl.petuum.com", "v1",
+                                 "", "adaptdljobs")
         self._cluster_expander = expander
         self._policy = PolluxPolicy()
+        # lock for the two corountines in run()
+        self.lock = asyncio.Lock()
```
aurickq, line 46: _lock for private variable

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,93 @@ class AdaptDLAllocator(object):
...
+        job = event["object"]
+        namespace = job["metadata"]["namespace"]
+        name = job["metadata"]["name"]
+
+        job = await self._objs_api.get_namespaced_custom_object(
+            "adaptdl.petuum.com", "v1", namespace, "adaptdljobs", name
+        )
+        # some other coroutine has handled this
+        if job.get("status", {}).get("allocation") or\
+                job.get("status", {}).get("group"):
```
aurickq, line 77: Need to check group is not None, since group==0 should trigger this condition too

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,93 @@ class AdaptDLAllocator(object):
...
+                        async with self.lock:
+                            await self._allocate_one(event)
+
+    async def _allocate_one(self, event):
+        # re-read the job , compared with the previous readjob
+        job = event["object"]
+        namespace = job["metadata"]["namespace"]
+        name = job["metadata"]["name"]
+
+        job = await self._objs_api.get_namespaced_custom_object(
```
aurickq, line 72: Need to handle 404 errors like is done in the controller, since the job may be deleted at any time. Maybe create an additional method in sched.utils to handle this for both places.

sched/adaptdl_sched/allocator.py
```diff
@@ -102,62 +168,71 @@ async def _find_nodes(self):
...
         job_list = await self._objs_api.list_namespaced_custom_object(
             "adaptdl.petuum.com", "v1", "", "adaptdljobs")
         job_infos = {}
         allocations = {}
+
         for job in job_list["items"]:
             if job.get("status", {}).get("phase") \
                     not in ["Pending", "Running", "Starting", "Stopping"]:
+                LOG.info("unkown status:", job,
+                         job.get("status", {}).get("phase"))
```
aurickq, line 223: Don't think we want to log anything here since this can print out a lot of stuff, e.g. for all completed jobs

sched/adaptdl_sched/policy/pollux.py
```diff
@@ -40,6 +40,39 @@ def __init__(self):
         self._min_util = 0.35
         self._max_util = 0.65
 
+    def _allocate_job(self, job_info, node_info_list):
```
aurickq, line 43: No underscore prefix for public method

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,93 @@ class AdaptDLAllocator(object):
...
+        name = job["metadata"]["name"]
+        # if this is a restarted job, skip i
+        LOG.info("detected an added job %s/%s.",
+                 namespace, name)
+        # parse the job infomation
+        job_info = {}
+        job_info[(namespace, name)] = self._get_job_info(job)
+
+        # find available nodes
+        node_info_list, _ = await self._find_nodes()
```
aurickq, line 90: Don't call it a list if it's not a list.

sched/adaptdl_sched/policy/pollux.py
```diff
@@ -40,6 +40,39 @@ def __init__(self):
...
+        """
+        A simple strategy that find the first available node for a new job.
+
+        Arguments:
+            job_info (dict): single element dictionary of the job info
+            node_info_list (dict): dict from node name to node_info
```
aurickq, line 49: Don't use list as a suffix if it's not a list

sched/adaptdl_sched/policy/pollux.py
```diff
@@ -40,6 +40,39 @@ def __init__(self):
         self._min_util = 0.35
         self._max_util = 0.65
 
+    def _allocate_job(self, job_info, node_info_list):
+        """
+        A simple strategy that find the first available node for a new job.
+
+        Arguments:
+            job_info (dict): single element dictionary of the job info
```
aurickq, line 48: Maybe better to take a job_key, job_info instead of a singleton dict so the semantics are clear from the arguments themselves

sched/adaptdl_sched/policy/pollux.py
```diff
@@ -40,6 +40,39 @@ def __init__(self):
...
+            node_info_list (dict): dict from node name to node_info
+
+        Returns:
+            list(str): allocation of the job,
+                e.g. [node name 0, node name 1, ...] if found available
+                     node, else None/
```
aurickq, line 54: Typo?

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,93 @@ class AdaptDLAllocator(object):
...
+
+        # get the node to allocate
+        new_allocation = self._policy._allocate_job(
+                            job_info, node_info_list)
```
aurickq, line 94: I think there is a bug here. The node resources contain the amount of resources not currently used by any AdaptDL job. So this will pass in more resources than are actually available, and cause the allocator to request more resources than there are available.

aurickq (changes requested):

sched/adaptdl_sched/allocator.py
YLJALDC: added individually. Since the one in controller returns different things.

sched/adaptdl_sched/allocator.py
```diff
@@ -38,27 +38,100 @@ class AdaptDLAllocator(object):
...
+                "adaptdl.petuum.com", "v1", namespace, "adaptdljobs", name
+            )
+        except kubernetes.client.rest.ApiException as exc:
+            if exc.status == 404:
+                return
+            raise  # unexpected
+
+        # some other coroutine has handled this
+        if job.get("status", {}).get("allocation") or\
+                job.get("status", {}).get("group") is not None:
```
aurickq, line 83: Hmm I feel this condition is still not very accurate, and a job could get sent through this loop multiple times (since each time the stream timeout occurs, ADDED events are generated again for every job). Maybe a better way is to let `status.allocations is None` indicate the job hasn't been seen by the allocator yet, and `status.allocations == []` indicate the job was seen by the allocator, but no nodes were allocated to it. We would only want to call `_allocate_one` for jobs with `status.allocations is None`. You'll probably also need to make sure this method will patch the job with `status.allocations == []` if it could not find an allocation for it.

aurickq (changes requested):

aurickq (approved):

PR #75, opened 2020-12-09
Add support for automatic mixed precision

### LINKED ISSUE #23: Support new PyTorch Automatic Mixed Precision
PyTorch 1.6 includes the stable release of automatic mixed precision (AMP). Make sure AdaptDL is fully compatible with this feature.

https://pytorch.org/docs/stable/notes/amp_examples.html

rmfan:
Adds support for automated mixed precision (https://pytorch.org/docs/stable/notes/amp_examples.html)

Resolves #23

Adds a new optional argument to `AdaptiveDataParallel`: `mp_scaler`, which will contain a user's `GradScaler` object. Defaults to `None` and is stored in `AdaptiveDataParallel` state object.


rmfan:
TODO: theres some times where the scaling constant is zero (which causes grad/var to become NaN and the scheduler to crash), need to investigate that further.

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -77,10 +77,11 @@ class AdaScale(object):
 
     .. _AdaScale: https://proceedings.icml.cc/static/paper_files/icml/2020/4682-Supplemental.pdf
     """  # noqa: E501
-    def __init__(self, optimizer, scale=None,
+    def __init__(self, optimizer, mp_scaler=None, scale=None,
```
aurickq, line 80: With the pluggable scaling rules PR, should figure out a better way to organize this code so that scaling rules don't need to be aware of mixed precision.

aurickq (changes requested):

adaptdl/adaptdl/torch/adascale.py
rmfan: The issue with having the Adascale code agnostic with mixed precision is that the mixed precision multiplier changes over time and we can only get it from the user's arbitrary `torch.cuda.amp.GradScaler` object. Unless we do something really weird with subclassing/monkeypatching `GradScaler`, the scaling rule will need to be aware of the `GradScaler` object. Alternatively, we could have an `update_amp_scaling` function in `AdaptiveDataParallel`, but I don't think thats any better.

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -217,9 +218,10 @@ def _backward_hook(self, idx, grad):
             self._local_sqr = torch.zeros(
                 (self._accumulation_steps, len(self._optimizer.param_groups)),
                 device=grad.device)
+        if self._mp_scaler is not None:
+            grad = grad / self._mp_scaler.get_scale()
```
aurickq, line 222: Last I checked, modifying the gradient here like this doesn't actually work..

rmfan: This is purely for the values stored in `self._local_sqr`. It's a rebinding of `grad` rather than an update.

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -227,14 +226,21 @@ def _final_callback(self):
...
+            mixed_precision_scale = 1.0
         for group in self._optimizer.param_groups:
             grads.append([])
             for param in group["params"]:
                 if param.grad is None:
                     grads[-1].append(None)
                     continue
-                param.grad.div_(self._accum_count)
-                grads[-1].append(param.grad.detach())
+                param.grad.div_(self._accumulation_steps)
```
aurickq, line 239: dividing by the wrong variable?

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -227,14 +226,21 @@ def _final_callback(self):
...
+                grad = param.grad.detach().float()
+                if (self._mp_scaler is not None):
```
aurickq, line 241: Don't think you need this check?

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -266,13 +276,19 @@ def _final_callback(self):
...
-            self.raw_grad_var = grad_var
-            theta = self._smoothing ** scale
-            self._update_avg('sqr_avg', grad_sqr, theta)
-            self._update_avg('var_avg', grad_var, theta)
+            # Note: mixed precision can result in nan/inf gradients,
+            # which propogate into our norm and variance estimates.
+            # Mixed precision autoscaling skips the skip where
+            # there are nan/inf, so we also skip the update here
+            if (np.all(np.isfinite(total_sqr))
+                    and np.all(np.isfinite(local_sqr))):
```
aurickq, line 284: This check might need to be a bit earlier, since we'd probably want to skip lines 250-276 as well (and definitely lines 260-262 and 274).

aurickq (changes requested):

adaptdl/adaptdl/torch/adascale.py
rmfan: Yea, I missed that change during the rebase

adaptdl/adaptdl/torch/adascale.py
rmfan: Pretty sure 
```
 if any(c != c or c in (float('inf'), -float('inf')) for c in check):
        print("AdaScale detected invalid gradient! Skipping step.")
        return
```
covers that as well. I think I can remove this final check.

adaptdl/adaptdl/torch/adascale.py
rmfan: removed

adaptdl/adaptdl/torch/adascale.py
aurickq: Sure, let's remove this then.

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -227,26 +226,44 @@ def _final_callback(self):
...
+        # Mixed precision autoscaling skips the skip where
+        # there are nan/inf, so we also skip the update here
+        grads_normsqr = _normsqr_groups(grads)
+        if not np.all(np.isfinite(grads_normsqr)):
             print("AdaScale detected invalid gradient! Skipping step.")
+            # Note: failing to zero here will accumulate the gradients,
+            # which will infinitely propogate the nan's/infs
+            # Also note: zeroing here will cause automatic mixed precision to
+            # never back off
+            self._accum_count = 0
```
rmfan, line 254: Note: I'd much rather have `_reset_accumulation()` not call `self._orig_optimizer_zero_grad()`, and then call `_reset_accumulation()` here.

aurickq: I think you can just call `_reset_accumulation` instead of `_orig_optimizer_zero_grad` in `zero_grad`. No need to reset any variables here, and that eliminates any chance that the gradients, `local_sqr`, and `accum_count` becomes de-synchronized due to programmer error.

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -246,7 +254,11 @@ def _final_callback(self):
         if count > 1:
             # Average local squared-norm samples.
             local_sqr = self._local_sqr.cpu().numpy() / count
-            total_sqr = _normsqr_groups(grads)
+            # Gradient is squared in local_sqr, so need to square the
+            # mixed precision scale as well
+            local_sqr = (local_sqr / mixed_precision_scale ** 2)
+            total_sqr = grads_normsqr
+            print(local_sqr, total_sqr)
```
aurickq, line 261: extraneous print

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -268,8 +280,8 @@ def _final_callback(self):
         if count > 1:
             grad_sqr = (count * total_sqr - local_sqr) / (count - 1)
             grad_var = (local_sqr - total_sqr) * scale / (count - 1)
-            self.raw_grad_sqr = grad_sqr
-            self.raw_grad_var = grad_var
+            grad_sqr = np.maximum(grad_sqr, 0.0)
+            grad_var = np.maximum(grad_var, 1e-6)
```
aurickq, line 284: We found that clipping post-smoothing performs better than clipping pre-smoothing, so should probably remove these two lines.

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -227,17 +226,26 @@ def _final_callback(self):
...
+                grads[-1].append(
+                    grad / mixed_precision_scale / self._accum_count)
+
+        # Note: mixed precision can result in nan/inf gradients,
+        # which propogate into our norm and variance estimates.
+        # Mixed precision autoscaling skips the skip where
+        # there are nan/inf, so we also skip the update here
+        grads_normsqr = _normsqr_groups(grads)
+        if not np.all(np.isfinite(grads_normsqr)):
             print("AdaScale detected invalid gradient! Skipping step.")
```
aurickq, line 249: Should probably be replaced with a `LOG.warn`.

aurickq (changes requested):

aurickq (approved):

PR #76, opened 2020-12-10
Add support for LEGW, Linear, and Sqrt scaling rules.

yukiontheiceberg:


adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -282,49 +416,24 @@ def _final_callback(self):
...
-        initial_lr = [pg["lr"] for pg in self._optimizer.param_groups]
-        for idx, param_group in enumerate(self._optimizer.param_groups):
-            grad_sqr = float(self._state["norm_avg"][idx])
-            grad_var = float(self._state["var_avg"][idx])
-            gain = (grad_var + grad_sqr) / (grad_var / self._scale + grad_sqr)
-            param_group["lr"] = gain * param_group["lr"]
-        self._optimizer_step(*args, **kwargs)
-        for lr, param_group in zip(initial_lr, self._optimizer.param_groups):
-            param_group["lr"] = lr
+class LinearScale(ScalingRuleBase):
```
rmfan, line 430: `get_progress` will need to be added to these classes too. I think having it return scale for this one should be fine (and sqrt(scale) for the `SqrtScale` class).

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -282,49 +416,24 @@ def _final_callback(self):
...
-    def step(self, *args, **kwargs):
-        """
-        Run one optimizer step using Adascale. Essentially just invokes
-        ``optimizer.step(*args, **kwargs)`` with a scaled learning rate.
+    def _calculate_scale_multiplier(self, pg_idx=None) -> float:
+        if pg_idx is None:
+            raise ValueError("Parameter index must be provided.")
+        grad_sqr = float(self._state["norm_avg"][pg_idx])
+        grad_var = float(self._state["var_avg"][pg_idx])
+        gain = (grad_var + grad_sqr) / (grad_var / self._scale + grad_sqr)
```
rmfan, line 426: I know this isn't implemented in the master branch, but this computation should be vectorizable. `self._state["norm_avg"]` and `self._state["var_avg"]` are numpy arrays, so we can first once produce a `gain` numpy array as well, and then index off of that.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -46,7 +48,190 @@ def _normsqr_groups(grads):
...
+        optimizer (torch.optim.Optimizer): Optimizer to apply this scaling rule
+        to.
+        scale (float): Scaling factor of the batch size, e.g. using a 10x
+            larger batch size (summed across all replicas) means a scale of
+            10. If None, defaults to ``num_replicas``.
+        num_replicas (int): Number of replicas for distributed training. If
+            None, defaults to ``torch.distributed.get_world_size()``.
+        patch_optimizer (bool): If True, monkey-patches the ``step`` method of
+            the optimizer with this scaling rule's ``step`` method.
+    """
```
aurickq, line 97: The overall logic is split between classes in a way that's not quite ideal currently, and probably makes the lr scaling class a bit more complicated than it needs to be. In particular, the norm and var estimation logic and get_progress logic should be outside of the scaling rule, because they should be the same regardless of the scaling rule, and only certain scaling rules need to access them (ie. adascale).

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -282,49 +444,38 @@ def _final_callback(self):
...
+                          self.scale, global_step)
+
+
+class SqrtScale(ScalingRuleBase):
+
+    def _calculate_scale_multiplier(self, pg_index=None) -> float:
+        return math.sqrt(self.scale)
+
+    def get_progress(self):
+        return math.sqrt(self.scale)
```
aurickq, line 477: I don't think this is correct mathematically. Progress should be calculated the same way as AdaScale in all cases.

adaptdl/adaptdl/torch/scaling_rules.py
aurickq: I don't think this is correct mathematically. Progress should be calculated the same way as AdaScale in all cases. Maybe factor out the norm/var estimation code and the progress calculation code out of the scaling rule classes.

adaptdl/adaptdl/torch/scaling_rules.py
yukiontheiceberg: I'm assuming that by "should be outside of the scaling rule" you mean that they need to be moved to the base class.
I understand that norm and var estimation logic remain the same regardless of the scaling rule, but before we move the logic to the base class I want to make sure that they make sense to be attributes attached to the base class --- if we simply add all invariant mathematical calculations to the base class it will end up being a huge Helper/Utility Class, in which case I think these logics might make more sense to be hosted somewhere else.

aurickq: Maybe a more maintainable/unit-testable way is to create a separate `GradientNoiseScale` class that calculates the norm/var statistics, and add it as a field of the AdaptiveDataParallel class? I don't think it should go into the base scaling rule class either, because it's not really a property of the scaling rule.

yukiontheiceberg: Tried to do that but it seems hard to completely decouple GradientNoiseScale from AdaScale. They are dependent on each other (GradientNoiseScale requires gradient accumulations and AdaScale requires stats calculations). So I made GradientNoiseScale a mixin class to add stats calculations to the ScalingRuleBase class.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -274,47 +314,83 @@ def _final_callback(self):
...
+    def _calculate_lr_factors(self, scale):
+        """Calculate factors to be applied to lr for each parameter group."""
+        lr_factors = []
+        for sqr, var, pg in zip(self._state["sqr_avg"], self._state["var_avg"],
+                                self._optimizer.param_groups):
+            sqr, var = max(float(sqr), 0.0), max(float(var), 1e-6)
+            lr_factors.append((var + sqr) / (var / scale + sqr))
+        return lr_factors
+
+    def calculate_lr_factor_estimation(self, scale):
```
rmfan, line 376: I think `_calculate_lr_factors` and `calculate_lr_factor_estimation` are named a bit too similar. Without reading their body's here, its difficult to figure out what they do and how they are different.

yukiontheiceberg: any suggestions?

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -274,47 +314,83 @@ def _final_callback(self):
...
+                loss.backward()
+                adascale.step()
+
+    .. _AdaScale: https://proceedings.icml.cc/static/paper_files/icml/2020/4682-Supplemental.pdf
+    """  # noqa: E501
+
+    def _calculate_lr_factors(self, scale):
+        """Calculate factors to be applied to lr for each parameter group."""
+        lr_factors = []
+        for sqr, var, pg in zip(self._state["sqr_avg"], self._state["var_avg"],
```
rmfan, line 370: You don't need `pg` here as its not used in the loop. Because `sqr` and `var` are numpy arrays, you can remove this loop and replace it with numpy math.

adaptdl/adaptdl/torch/scaling_rules.py
aurickq: The scaling rule class is still more complicated than it needs to be. Essentially, the new `AdaScale` class should just (roughly) be the logic in the `step` method of the old AdaScale class. Everything else can be factored out.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -91,9 +76,84 @@ def __init__(self, adp, optimizer, num_replicas=None,
...
+        """Output some useful metrics to TensorBoard."""
+        pass
+
+
+_Base = _ScalingRuleBase if typing.TYPE_CHECKING else object
+
+
+class _GradientNoiseScaleMixin(_Base):
+    """This mixin class makes _ScalingRuleBase able to track gradient related
+    stats."""
```
aurickq, line 149: Not sure this class makes sense to be a Mixin. Alternative is an object owned by AdaptiveDataParallel.

adaptdl/adaptdl/torch/scaling_rules.py
aurickq: I don't think this method really makes sense to be in the lr scaler class. Since it's just for progress tracking, which is independent of lr scaling.

yukiontheiceberg: This method is used to calculate the `Learning_Rate_Factor` scaler logged on Tensorboard, different from the lr_factor vector which is calculated per parameter group. It looks to me more like a single value estimation of the LR factor applied for this step. Not sure how it can be outside of scaling rules since it is calculated differently for each of them.

adaptdl/adaptdl/torch/scaling_rules.py
yukiontheiceberg: It is challenging to make it a stand-alone class owned by AdaptiveDataParallel mainly because GradientNoiseScale and ScalingRule depend on each other and is forming a cyclic dependency.
E.g. GradientNoiseScale needs ScalingRuleBase._accum_count https://github.com/petuum/adaptdl/blob/9b4ef6417038703c07d2c85b405935e0eb6a609c/adaptdl/adaptdl/torch/scaling_rules.py#L284
and ScalingRuleBase needs GradientNoiseScale.sqr_avg.
https://github.com/petuum/adaptdl/blob/9b4ef6417038703c07d2c85b405935e0eb6a609c/adaptdl/adaptdl/torch/scaling_rules.py#L211
I.e, we'll need to awkwardly create the two objects in a way like:
```python
gns = GradientNoiseScale(...)
sr = ScalingRule(...)
gns.sr = sr
sr.gns = gns
```
Unless we make accumulation steps related code a part of GradientNoiseScale, I don't see an easy way to completely remove the cyclic dependency.

PS. I simply moved all the callback functions to GradientNoiseScale which is not ideal because some accumulation steps related code is moved there altogether. But even if I further factor them out, the two examples above will remain there and cause the problem.

aurickq: Actually the gradient accumulation code _should_ be primarily in the gradient noise scale class. The lr scaler shouldn't need to know about gradient accumulation much at all.

The code snippet you posted looks fine, and can be hidden away in the AdaptiveDataParallel's `__init__` method, i.e.
```
self.scaling_rule = scaling_rule
self.scaling_rule.adp = self
```

adaptdl/adaptdl/torch/scaling_rules.py
aurickq: Oh, if it's just for the visualization, you can just take the average of the individual lr factors. No need to demand users implement custom logic for this.

aurickq (changes requested):

adaptdl/adaptdl/torch/parallel.py
```diff
@@ -35,20 +40,29 @@ class AdaptiveDataParallel(DistributedDataParallel):
...
     """
-    def __init__(self, model, optimizer, lr_scheduler=None,
-                 name="adaptdl-dataparallel", **kwargs):
+
+    available_scaling_rules = {cls.__name__: cls
+                               for cls in ScalingRuleBase.subclasses}
+
+    def __init__(self, model, optimizer,
+                 lr_scheduler=None, name="adaptdl-dataparallel",
+                 scaling_rule: Union[str, Type[SR]] = "AdaScale", **kwargs):
```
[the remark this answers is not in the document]
yukiontheiceberg: One side effect it creates is that the user instantiated object of ScalingRuleBase is not usable right away --- A GradientNoiseScale instance needs to be set by AdaptiveDataParallel before any method of ScalingRuleBase can be invoked, especially the `step` function. 
Moreover, the `patch_optimizer` method can no longer be invoked in `__init__` of either GradientNoiseScale or ScalingRuleBase
To keep the current behavior, the workflow equivalently becomes:
```python
# user code
optim = torch.optim.SGD(model, lr=0.001)
scaling_rule = AdaScale() # not usable
adp = AdaptiveDataParallel(model, optim, scaling_rule=scaling_rule)
# inside AdaptiveDataParallel
self.gns = GradientNoiseScale(self, optimizer)
self.scaling_rule=scaling_rule
self.scaling_rule.gns = self.gns
self.scaling_rule.patch_optimizer()
```

Looks fine to you?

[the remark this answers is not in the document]
aurickq: Looks fine, just `self.scaling_rule.gns = self.gns` should probably be `self.scaling_rule.adp = self`, since `adp.require_backward_gradient_sync` is needed by the patched step function to decide whether to take a step or not.

adaptdl/adaptdl/torch/parallel.py
```diff
@@ -59,9 +67,12 @@ def __init__(self, model, optimizer, lr_scheduler=None, mp_scaler=None,
         for param in model.parameters():
             param.register_hook(functools.partial(self._backward_hook, param))
 
-        # Setup for AdaScale, must be after registering backward hooks!
-        self.adascale = AdaScale(self, optimizer, mp_scaler=mp_scaler,
-                                 patch_optimizer=True)
+        # Setup for the scaling_rule, must be after registering backward hooks
+        # because some of them need to register their own backward hooks.
+        self.gns = GradientNoiseScale(self, optimizer, mp_scaler=mp_scaler)
+        self.scaling_rule = scaling_rule or AdaScale()
```
rmfan, line 73: I think this would be easier to use with a string to specify the class (and then create the object here), rather than forcing the user to create the object themselves

rmfan: Unless having the users be able to extend the scaling rules themselves is important

yukiontheiceberg: yeah, I was using string arg and then Aurick and I had a discussion here. The interface we currently have does allow for more flexibility. And for some scaling rules, like LEGW, we also need additional args to be passed in for initialization.
https://github.com/petuum/adaptdl/pull/76#discussion_r546293108

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -287,47 +261,176 @@ def _final_callback(self):
...
+        lr_factors = calculate_lr_factors_fn(scale)
+        for lr_factor, pg in zip(lr_factors, self.optimizer.param_groups):
+            pg["lr"] = lr_factor * pg["lr"]
         self._orig_optimizer_step(*args, **kwargs)
-        for lr, pg in zip(initial_lr, self._optimizer.param_groups):
+        for lr, pg in zip(initial_lr, self.optimizer.param_groups):
             pg["lr"] = lr
-        self._state["progress"] += self.gain(scale)
+        self.state["progress"] += self.gain(scale)
+        self._reset_accumulation()
```
rmfan, line 305: Remove this line: it can set the gradients to zero when the user doesn't expect them to be.

yukiontheiceberg: I might have missed it when I did the rebase. updating it

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -287,47 +261,175 @@ def _final_callback(self):
...
+                loss = ...
+                loss.backward()
+                adascale.step()
+    """
+    def __init__(self):
+        # instance of AdaptiveDataParallel, needs to be set before any of the
+        # methods can be used
+        self.adp = None
+
+    def calculate_lr_factors(self, scale):
```
aurickq, line 331: `scale_lr(self, scale)` to match research paper

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -287,47 +261,175 @@ def _final_callback(self):
...
+    Implements the AdaScale_ algorithm for scaling the learning rate for
+    distributed and large batch size training.
+
+    .. _AdaScale: https://proceedings.icml.cc/static/paper_files/icml/2020/4682-Supplemental.pdf
+    """  # noqa: E501
+
+    def calculate_lr_factors(self, scale):
+        """Calculate factors to be applied to lr for each parameter group."""
+        var = self.adp.gns.state["var_avg"]
+        sqr = self.adp.gns.state["sqr_avg"]
```
aurickq, line 369: Implement a `@property` instead of accessing gns state directly

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -287,47 +261,175 @@ def _final_callback(self):
...
+        var = np.maximum(var, 1e-6)
+        sqr = np.maximum(sqr,  0.0)
+        return (var + sqr) / (var / scale + sqr)
+
+
+class LinearScale(ScalingRuleBase):
+
+    def calculate_lr_factors(self, scale):
+        return np.full(len(self.adp.gns.optimizer.param_groups), scale)
```
aurickq, line 378: Better to make `optimizer` a `@property` of `AdaptiveDataParallel` than to access it as a field of `gns`

aurickq (changes requested):

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -156,14 +129,14 @@ def var_avg(self):
 
         Returns (float): Estimate of trace of the covariance.
         """
-        return float(np.sum(np.maximum(self._state["var_avg"], 1e-6)))
+        return float(np.sum(np.maximum(self.state["var_avg"], 1e-6)))
 
     def get_progress(self):
-        return self._state["progress"]
+        return self.state["progress"]
```
aurickq, line 135: Implement `@property` methods for instead of exposing the entire `state` as a public field.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -287,47 +261,175 @@ def _final_callback(self):
             self._update_avg('sqr_avg', grad_sqr, theta)
             self._update_avg('var_avg', grad_var, theta)
 
-    def step(self, *args, **kwargs):
+    def zero_grad(self, *args, **kwargs):
```
aurickq, line 264: I think the logic in `zero_grad`, `patch_optimizer`, and `run_step` should belong in `ScalingRuleBase` rather than here.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -53,42 +57,15 @@ def _normsqr_groups(grads):
...
-        num_replicas (int): Number of replicas for distributed training. If
-            None, defaults to ``torch.distributed.get_world_size()``.
-        patch_optimizer (bool): If True, monkey-patches the ``step`` method of
-            the optimizer with the AdaScale ``step`` method.
-
-    .. _AdaScale: https://proceedings.icml.cc/static/paper_files/icml/2020/4682-Supplemental.pdf
-    """  # noqa: E501
-    def __init__(self, adp, optimizer, mp_scaler=None, num_replicas=None,
-                 accum_scale=None, patch_optimizer=False):
+class GradientNoiseScale(object):
```
aurickq, line 60: This class should go to a different file.

adaptdl/adaptdl/torch/scaling_rules_test.py
```diff
@@ -37,11 +38,66 @@ def test_object():
...
+
+def test_scaling_rules_3():
+    """test SqrtScale lr factors"""
+    adp = Mock(require_backward_grad_sync=True)
+    opm = Mock(param_groups=[1, 0, 2, -1])
+    gns = Mock(optimizer=opm)
+    adp.gns = gns
+    sqrtscale = SqrtScale()
+    sqrtscale.adp = adp
+    np.testing.assert_equal(sqrtscale.calculate_lr_factors(9),
```
rmfan, line 78: Nit: should probably test these on more than one value. A for loop over a range should be fine I think.

adaptdl/adaptdl/torch/parallel.py
```diff
@@ -160,11 +173,11 @@ def to_tensorboard(self, writer, global_step, tag_prefix=""):
         if tag_prefix and not tag_prefix.endswith("/"):
             tag_prefix += "/"
         writer.add_scalar(tag_prefix + "Gradient_Norm_Sqr",
-                          self.adascale.sqr_avg(), global_step)
+                          self.gns.sqr_avg(), global_step)
         writer.add_scalar(tag_prefix + "Gradient_Variance",
-                          self.adascale.var_avg(), global_step)
+                          self.gns.var_avg(), global_step)
         writer.add_scalar(tag_prefix + "Learning_Rate_Factor",
-                          self._state.gain, global_step)
```
rmfan, line 167: I think there should still be the gain/efficiency value outputted to tensorboard

adaptdl/adaptdl/torch/scaling_rules_test.py
```diff
@@ -37,11 +38,66 @@ def test_object():
...
+               return_value=data_loader):
+        adp = Mock(require_backward_grad_sync=True)
+        opm = Mock(param_groups=[1, 0, 2, -1])
+        gns = Mock(optimizer=opm, get_progress=Mock(return_value=5))
+        adp.gns = gns
+        legwscale = LEGWScale(10, 1000)
+        legwscale.adp = adp
+        np.testing.assert_equal(legwscale.calculate_lr_factors(4),
+                                [0.025, 0.025, 0.025, 0.025])
+        gns = Mock(optimizer=opm, get_progress=Mock(return_value=400))
```
rmfan, line 95: You should probably also test this for a variety of progress values and with rescalings midway through.

adaptdl/adaptdl/torch/parallel.py
```diff
@@ -160,11 +173,11 @@ def to_tensorboard(self, writer, global_step, tag_prefix=""):
...
+                          self._state.lr_factor, global_step)
```
rmfan, line 180: Is this the only place where `lr_factor` is needed from state and isn't recomputed?

rmfan: If, so I think we can just recompute it here as well. If it needs to be fresh each time, then I don't think there is much point in having it in state.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -287,47 +261,175 @@ def _final_callback(self):
...
-            return
         scale = self._accum_scale * self._accum_count
-
-        initial_lr = [pg["lr"] for pg in self._optimizer.param_groups]
-        for sqr, var, pg in zip(self._state["sqr_avg"], self._state["var_avg"],
-                                self._optimizer.param_groups):
-            sqr, var = max(float(sqr), 0.0), max(float(var), 1e-6)
-            pg["lr"] = (var + sqr) / (var / scale + sqr) * pg["lr"]
+        initial_lr = [pg["lr"] for pg in self.optimizer.param_groups]
+        lr_factors = calculate_lr_factors_fn(scale)
```
aurickq, line 298: `scaled_lr = np.multiply(initial_lr, lr_factors)`, that way `calculate_lr_factors_fn` may also return a scalar instead of an array which simplifies several scaling rule implementations.

adaptdl/adaptdl/torch/gradient_noise_scale.py
```diff
@@ -1,34 +1,15 @@
...
 import torch.optim
+
 from torch.autograd import Variable
-from types import MethodType
 
 logging.basicConfig(level=logging.INFO)
 LOG = logging.getLogger(__name__)
 LOG.setLevel(logging.INFO)
 
-__all__ = ["AdaScale"]
```
rmfan, line 30: Change the `__all__` binding to `["GradientNoiseScale"]` rather than remove it so that the logging, `_average_groups`, etc aren't exposed.

adaptdl/adaptdl/torch/parallel.py
```diff
@@ -59,9 +67,13 @@ def __init__(self, model, optimizer, lr_scheduler=None, mp_scaler=None,
...
             param.register_hook(functools.partial(self._backward_hook, param))
 
-        # Setup for AdaScale, must be after registering backward hooks!
-        self.adascale = AdaScale(self, optimizer, mp_scaler=mp_scaler,
-                                 patch_optimizer=True)
+        # Setup for the scaling_rule, must be after registering backward hooks
+        # because some of them need to register their own backward hooks.
+        self.gns = GradientNoiseScale(self, optimizer, mp_scaler=mp_scaler)
+        self.scaling_rule = scaling_rule or AdaScale()
+        self.scaling_rule.adp = self
```
rmfan, line 74: Can this be moved to `register_optimizer` (and maybe rename that as `register_scaling_rule`)?

adaptdl/adaptdl/torch/gradient_noise_scale.py
```diff
@@ -114,32 +71,35 @@ def __init__(self, adp, optimizer, mp_scaler=None, num_replicas=None,
...
+        return self._accum_count
+
     def set_accum_scale(self, accum_scale):
         if not np.isclose(self._accum_scale, accum_scale):
             self._reset_accumulation()
             self._accum_scale = accum_scale
 
+    @property
+    def raw_sqr_avg(self):
+        return self._state["sqr_avg"]
```
aurickq, line 101: Recommend
```
view = self._state["sqr_avg"].view()
view.flags.writeable = False
return view
```
prevents `self._state["sqr_avg"]` from being accidentally mutated by the caller of this method. Same with `raw_var_avg`.

adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -0,0 +1,183 @@
...
+        self.adp = None
+        self._optimizer = None
+        self._orig_optimizer_step = None
+
+    def scale_lr(self, scale):
+        raise NotImplementedError
+
+    def zero_grad(self, *args, **kwargs):
+        if self.adp.gns._should_zero_grad:
+            self.adp.gns._reset_accumulation(*args, **kwargs)
```
aurickq, line 60: `_reset_accumulation` should be made public, and `_should_zero_grad` accessed through a `@property def should_zero_grad`

aurickq (changes requested):

adaptdl/adaptdl/torch/parallel.py
rmfan: nvm, needs the current scale, which requires state

adaptdl/adaptdl/torch/scaling_rules_test.py
```diff
@@ -199,8 +285,10 @@ def nan_objective(tensor):
     params = torch.autograd.Variable(params_t, requires_grad=True)
     sgd = torch.optim.SGD([params], lr=0.1)
     adp = Mock(require_backward_grad_sync=True)
-    ada = adascale.AdaScale(adp, sgd, accum_scale=1.0, num_replicas=1,
-                            patch_optimizer=True)
+    gns = GradientNoiseScale(adp, sgd, accum_scale=1.0, num_replicas=1)
```
rmfan, line 288: I think test_nan() should be moved into a gradient_noise_scale_test.py file, with the AdaScale object removed.

adaptdl/adaptdl/torch/scaling_rules_test.py
```diff
@@ -18,17 +18,19 @@
...
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
-import adaptdl.torch.adascale as adascale
+from adaptdl.torch.gradient_noise_scale import GradientNoiseScale
+from adaptdl.torch.scaling_rules import AdaScale, LinearScale,\
+     LEGWScale, SqrtScale
 
 
 def test_object():
```
rmfan, line 28: This test can also probably be moved to `gradient_noise_scale_test.py`

rmfan (approved):

aurickq (approved):

PR #79, opened 2020-12-13
Adaptive batch size updates from research codebase

aurickq:


adaptdl/adaptdl/torch/adascale.py
```diff
@@ -293,22 +279,25 @@ def step(self, *args, **kwargs):
...
-        for lr, param_group in zip(initial_lr, self._optimizer.param_groups):
-            param_group["lr"] = lr
+        for sqr, var, pg in zip(self._state["sqr_avg"], self._state["var_avg"],
+                                self._optimizer.param_groups):
+            sqr, var = max(float(sqr), 0.0), max(float(var), 1e-6)
+            pg["lr"] = (var + sqr) / (var / scale + sqr) * pg["lr"]
+        if self._optimizer_step is None:
+            self._optimizer.step(*args, **kwargs)
+        else:
+            self._optimizer_step(*args, **kwargs)
```
rmfan, line 293: Nit/style: `self._optimizer_step` is too close to `self._optimizer.step`

adaptdl/adaptdl/torch/_metrics.py
```diff
@@ -105,24 +103,27 @@ def get_goodput_fn():
...
-             for val in values])
-    compute_time = step_time - sync_time
-    accumulation_time = np.where(
-        accumulation_steps > 0, accumulation_time, compute_time)
-    state.perf_params = fit_perf_params(
-        num_nodes, num_replicas, local_bsz, accumulation_steps,
-        step_time, compute_time, accumulation_time)
+    profile = {k: v for k, v in state.profile.items() if v.get("optim_count")}
+    # Convert profile into numpy arrays.
+    num_nodes, num_replicas, atomic_bsz = (np.array(k) for k in zip(*profile))
```
rmfan, line 108: I think the fact that `*profile` only takes the keys is a bit esoteric. Could you add that fact to the comment above?

adaptdl/adaptdl/torch/_metrics.py
```diff
@@ -105,24 +103,27 @@ def get_goodput_fn():
...
+    accum_step_time = np.array([v.get("accum_step_time", 0.0)
+                                for v in profile.values()])
+    accum_count = np.array([v.get("accum_count", 0) for v in profile.values()])
+    optim_step_time = np.array([v.get("optim_step_time", 0.0)
+                                for v in profile.values()])
+    optim_sync_time = np.array([v.get("optim_sync_time", 0.0)
+                                for v in profile.values()])
+    optim_count = np.array([v.get("optim_count", 0) for v in profile.values()])
+    assert np.all(optim_count > 0)
```
rmfan, line 117: Is a hard error the right call here? This will kill the adaptdljob entirely right?

adaptdl/adaptdl/torch/adascale_test.py
```diff
@@ -103,25 +111,22 @@ def rosenbrock(x, y):
...
-    assert(params_t[0]['params'][0].allclose(torch.tensor([1.0]),
-                                             atol=0.01)
-           and params_t[1]['params'][0].allclose(torch.tensor([1.0]),
-                                                 atol=0.01))
+        if params_t[0]['params'][0].allclose(torch.tensor([1.0]), atol=0.01) \
+                and params_t[1]['params'][0].allclose(torch.tensor([1.0]),
+                                                      atol=0.01):
+            break
+    else:
+        assert False, params_t
```
rmfan, line 129: Can you add a print/log that this error is caused by the optimizer not converging in the proper number of steps?

adaptdl/adaptdl/torch/adascale.py
```diff
@@ -77,96 +81,72 @@ class AdaScale(object):
...
 
         self._smoothing = 0.999
 
+        self.raw_grad_sqr = np.ones(len(optimizer.param_groups))
+        self.raw_grad_var = np.zeros(len(optimizer.param_groups))
+
+        self._reset_accumulation()
+
+    def _reset_accumulation(self):
+        if self._optimizer_zero_grad is None:
```
rmfan: I'm not a huge fan of this change to be honest. I don't think its a good idea to fully automate zeroing of gradients and to force zeroing during the `step` call.

adaptdl/adaptdl/torch/_metrics.py
aurickq: It should never happen because of line 106. Add the assert here to catch future programming errors

rmfan (approved):
LGTM

Issue #70, opened 2020-11-24
Ensure saving checkpoints is atomic


aurickq:
Currently, checkpoints can be corrupted if the Kubernetes termination grace period is exceeded and the pod is forcibly killed, which causes the job to crash after restart. We can make sure that the checkpoints are consistent by saving to a different directory and renaming it after the full checkpoint it saved.

PR #81, opened 2020-12-17
Save Checkpoint File Atomically

hao-howard-zhang:
This PR:

1. saves checkpoint atomically by writing to a temporary file, and uses `os.rename()` to guarantee atomicity.
2. removes unused chart template: `cluster-bindings.yaml` and `permissions.yaml`

Tested on cluster by running `bash tests/long-workload/resnet18-cifar10-elastic.sh` and the result on tensorboard looks correct.

Closing https://github.com/petuum/adaptdl/issues/70


rmfan (approved):

adaptdl/adaptdl/checkpoint.py
```diff
@@ -110,10 +110,17 @@ def save_state(state, sync=True):
...
-            with open(os.path.join(checkpoint_path(), name), "wb") as f:
+            ckpt_file = os.path.join(checkpoint_path(), name)
+            tmp_ckpt_file = os.path.join(checkpoint_path(), tmp_name)
+
+            with open(tmp_ckpt_file, "wb") as f:
                 state.save(f)
 
+            # Save checkpoint atomically to avoid file corruption
+            os.rename(tmp_ckpt_file, ckpt_file)
+
```
aurickq, line 123: I think this approach still has an issue when the process is killed in-between saving two state objects. In that case the overall checkpoint may consist of only partially-updated states. Better to create a sub-directory for each checkpoint I think, and delete the previous sub-directories.

hao-howard-zhang: Makes sense. Checkpoints are saved by invoking `save_all_states()`, not individual `save()` function. I missed that. Thanks for pointing it out!

hao-howard-zhang: Now the `save_state()` will write to a temp folder (`_checkpoint`). Once all states are saved, the temp folder will be renamed to `checkpoint`. Please review again.

aurickq (approved):

adaptdl/adaptdl/checkpoint.py
```diff
@@ -95,24 +99,41 @@ def save_all_states():
...
+    # Prevent corrupting original state files in case the process got killed
+    # during state file writing.
+    if replica_rank() == 0 and checkpoint_path() is not None:
+        ckpt_dir = os.path.join(checkpoint_path(), CKPT_DIR_NAME)
+        if os.path.isdir(ckpt_dir):
+            shutil.rmtree(ckpt_dir)
+
+        os.makedirs(ckpt_dir, exist_ok=True)
+        ckpt_tmp_dir = os.path.join(checkpoint_path(), CKPT_TMP_DIR_NAME)
+        os.rename(ckpt_tmp_dir, ckpt_dir)  # atomic
```
aurickq, line 111: The process is still not quite atomic. e.g. say the pod gets killed during the `rmtree` (which can take a non-trivial amount of time when using a DFS). A better way is to save to a unique directory each time, e.g. `checkpoint-{n}` where `n = adaptdl.env.num_restarts()`, and use a symlink to point to the latest directory. Then, delete the old checkpoints only after the symlink is written.

aurickq (changes requested):

adaptdl/adaptdl/checkpoint.py
hao-howard-zhang: You are right. I will make the change as suggested. Thanks!

adaptdl/adaptdl/checkpoint.py
```diff
@@ -95,24 +107,45 @@ def save_all_states():
     for state in _STATES_TO_NAMES:
         save_state(state)
 
+    # Prevent corrupting original state files in case the process got killed
+    # during state file writing.
+    if replica_rank() == 0 and checkpoint_path() is not None:
+        tmp_ckpt_dir = _get_tmp_ckpt_dir()
+        ckpt_dir = os.path.join(checkpoint_path(),
+                                f"{CKPT_DIR_PREFIX}{num_restarts()}")
+        os.makedirs(ckpt_dir, exist_ok=True)
```
aurickq, line 116: Don't think you need to create this dir

adaptdl/adaptdl/checkpoint.py
```diff
@@ -95,24 +107,45 @@ def save_all_states():
...
+        os.rename(tmp_ckpt_dir, ckpt_dir)  # atomic
+
+        for old_ckpt_dir in os.listdir(checkpoint_path()):
+            if (
+                os.path.isdir(old_ckpt_dir) and
+                old_ckpt_dir.startswith(CKPT_DIR_PREFIX) and
+                old_ckpt_dir != ckpt_dir
+            ):
+                shutil.rmtree(old_ckpt_dir)
```
aurickq, line 125: `os.listdir` will output just the directory name without the path, so you'll need to prepend the path to it.

adaptdl/adaptdl/checkpoint.py
```diff
@@ -130,10 +163,15 @@ def load_state(state):
...
-    try:
-        name = _STATES_TO_NAMES[state]
-        with open(os.path.join(checkpoint_path(), name), "rb") as f:
-            state.load(f)
-        return True
-    except FileNotFoundError:
+
+    name = _STATES_TO_NAMES[state]
+    ckpt_dir = os.path.join(checkpoint_path(),
+                            f"{CKPT_DIR_PREFIX}{num_restarts() - 1}")
```
aurickq, line 169: This makes it if the previous checkpoint failed, it will restart from the beginning. Probably better to find the latest checkpoint, even if it's not the immediate previous restart. Can output a warning if it's not from the previous restart since that can be useful debugging info.

aurickq (changes requested):

adaptdl/adaptdl/checkpoint.py
hao-howard-zhang: Thanks. Updated

adaptdl/adaptdl/checkpoint.py
hao-howard-zhang: Good catch. I have updated the code. I will do a quick integration test with resnet.

hao-howard-zhang: Ran `resnet18-cifar10-elastic.sh` and the result looks good.

aurickq (approved):

PR #83, opened 2020-12-27
Version check between AdaptDL Trainer Lib and Scheduler

ZeyaWang:
Add a version check between AdaptDL trainer and scheduler


adaptdl/adaptdl/torch/__init__.py
```diff
@@ -37,6 +39,15 @@
 LOG.setLevel(logging.INFO)
 
 
+def version_check(version, lib):
+    if re.match("[0-9].[0-9].[0-9]", version) and \
+            version != "0.0.0":
+        return True
+    else:
+        LOG.info("adaptdl version of {} is a customer version".format(lib))
```
rmfan, line 47: Nit: I'm not sure customer is the correct word here -- maybe user or custom?

ZeyaWang: Yeah, it is a typo, just change it to "user".

sched/adaptdl_sched/controller.py
```diff
@@ -391,6 +393,10 @@ async def _create_pod(self, job_metadata, pod_template,
                 "name": "ADAPTDL_SUPERVISOR_URL",
                 "value": config.get_supervisor_url(),
             })
+            container["env"].append({
+                "name": "ADAPTDL_VERSION",
+                "value": config.get_adaptdl_version(),
+            })
```
aurickq, line 399: `ADAPTDL_SCHED_VERSION`.

Also, if you're passing the version as an env var then you shouldn't need the changes to the supervisor. You can simply read the env var on the trainer-side.

adaptdl/adaptdl/torch/__init__.py
```diff
@@ -47,9 +58,17 @@ def init_process_group(backend):
...
         response.raise_for_status()
-        master_addr = response.json()[0]
+        master_addr = response.json()['pod_ip_list'][0]
+        sched_version = response.json()['version']
+        trainer_version = pkg_resources.get_distribution("adaptdl").version
+        if version_check(sched_version, lib="sched") and \
+                version_check(trainer_version, lib="trainer"):
+            if trainer_version != sched_version:
+                raise Exception('The adaptdl version between trainer \
+                    and scheduler should be same')
```
aurickq, line 68: Checking for exact version match is too strict, since that requires every adaptdljob to use the exact same version of the adaptdl library. Instead the version checking should defer to a function that determines compatibility, i.e. whenever we make a breaking change, we add another condition to that function.

aurickq (changes requested):

adaptdl/adaptdl/torch/__init__.py
ZeyaWang: Sure, checking compatibility instead of strict version checking makes much more sense. Do you have any suggestions about the design for detecting the breaking change?

ZeyaWang: After discussion with Omkar and Richard, we agree to follow semantic versioning (https://semver.org/) to increment the version so the version checking of compatibility will be implemented in terms of the major version number.

adaptdl/adaptdl/torch/__init__.py
```diff
@@ -48,8 +60,18 @@ def init_process_group(backend):
                 break
         response.raise_for_status()
         master_addr = response.json()[0]
+        sched_version = os.environ.get("ADAPTDL_VERSION")
```
aurickq, line 63: Rename env var to `ADAPTDL_SCHED_VERSION` to make clear this is the scheduler's version.

ZeyaWang: Updated

adaptdl/adaptdl/torch/__init__.py
```diff
@@ -48,8 +60,18 @@ def init_process_group(backend):
                 break
         response.raise_for_status()
         master_addr = response.json()[0]
+        sched_version = os.environ.get("ADAPTDL_SCHED_VERSION")
```
aurickq, line 63: Add this as a function in `env.py`

adaptdl/adaptdl/torch/__init__.py
```diff
@@ -48,8 +60,18 @@ def init_process_group(backend):
...
+        trainer_version = pkg_resources.get_distribution("adaptdl").version
+        if version_check(sched_version, lib="sched") and \
+                version_check(trainer_version, lib="trainer"):
+            trainer_ver_maj = semver.VersionInfo.parse(trainer_version).major
+            sched_ver_maj = semver.VersionInfo.parse(sched_version).major
+            if trainer_ver_maj != sched_ver_maj:
+                raise Exception('The adaptdl version between trainer'
+                                ' and scheduler should be compatible')
```
aurickq, line 71: Provide more info here, e.g. "adaptdl version x.x.x is incompatible with scheduler version x.x.x"

adaptdl/adaptdl/torch/__init__.py
```diff
@@ -37,6 +40,15 @@
 LOG.setLevel(logging.INFO)
 
 
+def version_check(version, lib):
+    if semver.VersionInfo.isvalid(version) and \
+            version != "0.0.0":
+        return True
+    else:
+        LOG.info("adaptdl version of {} is a user version".format(lib))
```
aurickq, line 48: Not sure what a "user version" is, do you need to print anything here?

aurickq (changes requested):

aurickq (dismissed):

aurickq (approved):

PR #91, opened 2021-02-11
AdamScale Support

sangkeun00:
AdamScale is a variant of AdaScale that adaptively adjusts learning rate.

### DIFF
--- a/adaptdl/adaptdl/torch/gradient_noise_scale.py
+++ b/adaptdl/adaptdl/torch/gradient_noise_scale.py
@@ -32,7 +32,7 @@ def _normsqr_groups(grads, pinvs):
     ret = []
     for group, pinv_group in zip(grads, pinvs):
         normsqr = [(g / pinv).pow(2).sum(dtype=torch.float64)
-                    for g, pinv in zip(group, pinv_group) if g is not None]
+                   for g, pinv in zip(group, pinv_group) if g is not None]
         ret.append(sum(normsqr).item() if normsqr else 0.0)
     return np.array(ret)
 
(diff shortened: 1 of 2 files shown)


adaptdl/adaptdl/torch/scaling_rules.py
```diff
@@ -125,6 +125,41 @@ def scale_lr(self, scale):
...
+        sqr = np.maximum(sqr,  0.0)
+        return np.power((var + sqr) / (var / scale + sqr), power)
+
+    def step(self, *args, **kwargs):
+        if not self.adp:
+            raise ValueError("AdaptiveDataParallel instance is not set!")
+        if not self.adp.require_backward_grad_sync:
+            return
+        scale = self.adp.gns.accum_scale * self.adp.gns.accum_count
+        if self.adp.gns._is_adam and \
```
rmfan, line 147: Is there ever a case when the AdamScale class is used and the gns doesnt have the adam flag set?

adaptdl/adaptdl/torch/gradient_noise_scale.py
```diff
@@ -53,6 +54,9 @@ def __init__(self, adp, optimizer,
                               else torch.distributed.get_world_size())
         self._accum_scale = accum_scale or self._num_replicas
         self._prev_grads = None
+        self._is_adam = optimizer.__class__.__name__ in ['Adam', 'AdamW']
```
rmfan, line 57: I think it would be cleaner to use a new subclass with a different `_calculate_preconditioner` function, where the base `GradientNoiseScale._calculate_preconditioner` returns the identity preconditioner.

rmfan (changes requested):

rmfan:
Merged into master as part of #95

PR #101, opened 2021-10-06
Adaptive Tune Trial Scheduler

odp:
This PR implements Adaptive Tune Trial Scheduler which effectively ports AdaptDL over Ray. The AdaptDL Trial Scheduler can be used like any other Tune scheduler with any Tune `DistributedTrainable`.  The full proposal can be found at https://github.com/ray-project/ray/issues/17054 

The implementation consists of below components:

1. `AdaptDLScheduler`
This is the user-exposed class that implements the elastic scheduling functionality from AdaptDL for Tune trials. It invokes the AdaptDL allocator at a fixed frequency counted in the number of epochs observed by the scheduler (which can be controlled by the parameter `allocator_invoke_freq`). The job of the scheduler is to find the optimal resources for each trial of the experiment with the help of the `AdaptDLAllocator` and then applying those changes to the running trials seamlessly. As a result, individual trials can expand or contract throughout their lifetimes, by changing the number of replicas they run with, based on their performance characteristics and available Ray cluster resources. We maintain the model and dataloader state during the rescaling events with the help of checkpoints managed by the `AdaptDLTrial`s.

2. `AdaptDLJobMixin`
This is an interface between Tune trials and AdaptDL jobs (as seen by the `AdaptDLAllocator`). The mixin presents a clean interface to `AdaptDLAllocator` which is responsible for fairly distributing available cluster resources between all the trials. It primarily holds the current allocation and the `JobInfo` structure for the `AdaptDLTrial`. This is used by both the `AdaptDLAllocator` and `AdaptDLScheduler` to make decisions about trials and execute them.

3. `AdaptDLAllocator`
Uses the interface implemented by the mixin to generate an allocation plan for the jobs/trials. This is a wrapper over the job allocator from AdaptDL.

4. `AdaptDLTrial(AdaptDLJobMixin, Tune.Trial)`
This combines the interface mixin with Tune trials. It mainly implements two methods that are used by the scheduler, namely `create_from` which creates or rescales the `AdaptDLTrial` and `pause` which pauses the `AdaptDLTrial` and releases all resources allocated to it. This presents a simple interface to the `AdaptDLScheduler` to control `AdaptDLTrial`s. 

5. `AdaptDLTrainableCreator`
This module implements the function that creates AdaptDL `DistributedTrainable`  from the user-provided training function (see examples). Trainables are classes which comply with the Tune specification of a `Trainable` and methods of which are invoked by Tune to control the execution of trials created from them. In our case, because the trainables are distributed, they end up spawning more replica actors which are responsible for running the training loop. It is the responsibility of the `AdaptDLAllocator` to provide resources for the entire ensemble. 

Depends on https://github.com/ray-project/ray/pull/19349

### DIFF
--- a/ray/adaptdl_ray/tune/adaptdl_trainable.py
+++ b/ray/adaptdl_ray/tune/adaptdl_trainable.py
@@ -1,0 +1,70 @@
+# Copyright 2021 Petuum, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from typing import Callable, Dict, Generator, Optional, Type
+from unittest.mock import patch
+from datetime import timedelta
+
+import ray
+from ray.tune.resources import Resources
+from ray.tune.registry import get_trainable_cls, register_trainable
+from ray.tune.integration.torch import _TorchTrainable
+from ray.util.sgd.torch.constants import NCCL_TIMEOUT_S
+
+import adaptdl_ray.tune.adaptdl_patch as P 
+
+
+def AdaptDLTrainableCreator(func: Callable,
+                            num_workers: int = 1,
+                            group: int = 0,
+                            num_cpus_per_worker: int = 1,
+                            num_gpus_per_worker: int = 0,
+                            num_workers_per_host: Optional[int] = None,
+                            backend: str = "gloo",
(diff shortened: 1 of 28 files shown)


ray/adaptdl_ray/tune/adaptdl_trainable.py
```diff
@@ -0,0 +1,137 @@
...
+        torch.nn.Linear(D_in, H),
+        torch.nn.ReLU(),
+        torch.nn.Linear(H, D_out),
+    )
+    optimizer = optim.SGD(model.parameters(), lr=0.1)
+
+    model = adl.AdaptiveDataParallel(model, optimizer)
+
+    loss = torch.Tensor([0.0])
+    for epoch in adl.remaining_epochs_until(config.get("epochs", 10)):
```
rmfan, line 129: With relatively expensive tests like this, we should probably have a separate job for testing the /ray/adaptdl_ray directory so it can run in parallel with the k8s and client tests. Those take enough time already.

rmfan: This isn't a blocking comment, btw

aurickq (approved):

aurickq (approved):

PR #102, opened 2021-10-06
Add support to run an adaptdl job on a ray aws cluster

rmfan:
Adds a python script/command line utility that allows a user to submit an adaptdl job on an existing ray-aws cluster with the following properties:

- Spot instances termination is supported.
- Variable cluster size is supported
- AdaptDL user code doesn't need to change from the standard AdaptDL user code
- Only one AdaptDL job per ray cluster
- User code can be local (doesn't need to be in the ray worker docker images, etc)


ray/adaptdl_ray/aws/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM rayproject/ray-ml:1.7.0-py38-gpu
+RUN sudo apt update && sudo apt install -y awscli
+RUN python3 -m pip install adaptdl adaptdl_sched tensorboard boto3 aiohttp requests
+RUN wget https://aws-s3-plugin.s3.us-west-2.amazonaws.com/binaries/0.0.1/1c3e69e/awsio-0.0.1-cp38-cp38-manylinux1_x86_64.whl && python3 -m pip install awsio-0.0.1-cp38-cp38-manylinux1_x86_64.whl && rm -f awsio-0.0.1-cp38-cp38-manylinux1_x86_64.whl
+RUN python3 -m pip install pymoo==0.4.0
```
odp, line 5: May be this needs to change pymoo>=0.5.0 (where adaptdl master is at)

ray/adaptdl_ray/adaptdl/utils.py
```diff
@@ -34,15 +35,19 @@ def pgf_to_allocation(pgf) -> List[str]:
...
         if config.default_device() == "GPU":
-            # As per Ray, We need equal amount of CPUs if there are GPUs in
-            # this bundle
-            resources["CPU"] = device_count
+            resources_per_node["GPU"] = 1.0
+
+    def _construct_bundle(node, number_of_instances):
+        resources = deepcopy(resources_per_node)
+        resources["CPU"] *= number_of_instances
+        resources["GPU"] *= number_of_instances
```
odp, line 48: could produce KeyError if `default_device()` is `"CPU"`

ray/adaptdl_ray/adaptdl/utils.py
```diff
@@ -34,15 +35,19 @@ def pgf_to_allocation(pgf) -> List[str]:
     return allocs
 
 
-def allocation_to_pgf(alloc: List[str]):
+def allocation_to_pgf(alloc: List[str], resources_per_node=None):
```
odp, line 38: Instead of `None`, maybe initialize with `{"CPU" = 1.0}`? so the `if not` block can be removed.

odp (changes requested):

ray/adaptdl_ray/aws/Dockerfile
rmfan: Don't need this file anymore

ray/adaptdl_ray/adaptdl/utils.py
rmfan: Python doesn't do object as defaults well (there is only one object per that function globally)

aurickq (approved):

PR #115, opened 2022-04-04
Stage1

Xuezhi-Liang:
Unknown

Notes: 

1. The first step is to create the Context class and start to move the core AdaptDL mathematics into that class. 

2. The AdaptiveDataLoader should get refactored to use the Context class for those AdaptDL logic to prevent duplication of logic. 

3. But the changes to AdaptiveDataLoader are purely implementation changes, users of the class shouldn't need to see any difference.

### DIFF
--- a/adaptdl/adaptdl/torch/context.py
+++ b/adaptdl/adaptdl/torch/context.py
@@ -1,0 +1,153 @@
+# Copyright 2020 Petuum, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import collections
+import math
+import adaptdl.checkpoint
+import adaptdl.collective
+import adaptdl.env
+from adaptdl.torch._metrics import get_goodput_fn
+import adaptdl.torch.data as data
+import numpy as np
+
+class Context(object):
+    """
+    This class provides context tool to get AdaptDL-suggest parameters,
+    such as batch_size, accum_steps and lr_scale.
+    """
+
+    def __init__(self, batch_size=32):
+        # Autoscale batch size fields.
+        self._speedup_threshold = 1.05
+        self.adapt_batch_size = None
+        self.adapt_accum_steps = None
+        self.adapt_lr_scale = None
+
+        self._max_batch_size = None
+        self._local_bsz_bounds = None
+        # Create and load state.
+        self._state = data._AdaptiveDataLoaderState()
+        adaptdl.checkpoint.load_state(self._state)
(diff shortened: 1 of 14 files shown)


adaptdl/adaptdl/torch/context.py
```diff
@@ -0,0 +1,98 @@
...
+from adaptdl.torch.scaling_rules import ScalingRuleBase
+
+class AdaptiveDLContext(object):
+    """
+    This class provides context tool to get AdaptDL-suggest parameters,
+    such as batch_size, accum_steps and lr_scale.
+    """
+
+    def __init__(self, batch_size):
+        self._elastic = adaptdl.torch.data.AdaptiveDataLoaderHelper(batch_size)
```
aurickq, line 31: Should make sure the `AdaptiveDLContext` class does not depend on the `AdaptiveDataLoaderHelper` class (or anything related to PyTorch).

adaptdl/adaptdl/torch/context.py
```diff
@@ -0,0 +1,98 @@
...
+                adaptdl.env.num_nodes(), adaptdl.env.num_replicas(),
+                self._elastic.current_local_bsz, self._elastic.accumulation_steps)
+            # use only if speedup is significant
+            speedup = suggest_goodput / max(current_goodput, 1e-8)
+            if speedup > self._speedup_threshold:
+                self._elastic._state.current_local_bsz = atomic_bsz
+                self._elastic._state.accumulation_steps = accum_steps
+        self._elastic._state.current_local_bsz, self._elastic._state.accumulation_steps = \
+            adaptdl.collective.broadcast((self._elastic._state.current_local_bsz,
+                                          self._elastic._state.accumulation_steps))
```
aurickq, line 89: Shouldn't do any cross-replica synchronization from a call to the `AdaptiveDLContext` class. The synchronization should happen outside, from the callsite.

adaptdl/adaptdl/torch/context.py
```diff
@@ -0,0 +1,98 @@
...
+        return self._elastic.current_local_bsz, self._elastic._state.current_local_bsz, self._elastic._state.accumulation_steps
+
+    @property
+    def training(self):
+        return self._elastic.training
+
+    def to_tensorboard(self, writer, global_step, tag_prefix=""):
+        self._elastic.to_tensorboard(writer, global_step, tag_prefix)
```
aurickq, line 97: Tensorboard isn't core AdaptDL logic so shouldn't be included in this class.

adaptdl/adaptdl/torch/context.py
```diff
@@ -0,0 +1,98 @@
...
+# limitations under the License.
+
+import collections
+import math
+import adaptdl.checkpoint
+import adaptdl.collective
+import adaptdl.env
+from adaptdl.torch._metrics import get_goodput_fn
+import adaptdl.torch.data
+from adaptdl.torch.scaling_rules import ScalingRuleBase
```
aurickq, line 22: Should make sure this module doesn't depend on anything from PyTorch

adaptdl/adaptdl/torch/context.py
```diff
@@ -0,0 +1,98 @@
...
+
+class AdaptiveDLContext(object):
```
aurickq, line 24: Maybe rename it to just `Context` so usage of it can be simply `adaptdl.Context`

aurickq (changes requested):

Xuezhi-Liang:
close via a new PR #123

PR #123, opened 2022-05-31
Stage1.5

Xuezhi-Liang:


adaptdl/adaptdl/torch/data.py
```diff
@@ -120,6 +120,24 @@ def current_dataloader():
     return AdaptiveDataLoaderHelper._current
 
 
+Context_obj = None
+def context_initialize(batch_size):
+    """
+    Initialize this module, must be invoked before calling any other functions.
+    This function will block until it has been invoked from all replicas.
```
odp, line 127: How's this enforced?

VincentYaoMBZUAI: Dear Omkar, Thank you for asking. As we were making the Context global, the Context was firstly initialized in init_process_group as Context_obj following Aurick's suggestion. All the subsequent processes in terms of Context will be using the Context_obj instead. So the initialize process was enforced in the very beginning.

adaptdl/adaptdl/torch/__init__.py
```diff
@@ -119,6 +120,9 @@ def init_process_group(backend,
                                   rank,
                                   world_size)
 
+    # Initialize Context module.
+    adaptdl.torch.data.context_initialize(batch_size=32)
```
odp, line 124: How do you plan to make `batch_size` available in `init_process_group`?

VincentYaoMBZUAI: > How do you plan to make `batch_size` available in `init_process_group`?

We were just giving it a default value here for the global Context, the batch_size could be replaced by users when they use it. But thanks for point it out that we can also remove the default value from here, which would be given from the definition of class Context directly. Further, users are still able to give their input value and the default value will be replaced then. 
Please also have a check with our new commit to review the modification.
Many thanks.

adaptdl/adaptdl/torch/data.py
```diff
@@ -355,7 +318,7 @@ def context(self):
 
     @property
     def current_batch_size(self):
-        return (self.current_local_bsz * (self.accumulation_steps + 1) *
+        return (self._context.get_batch_size() * (self._context.get_accum_steps() + 1) *
```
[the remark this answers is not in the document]
VincentYaoMBZUAI: Thank you Omkar, Please have a check on the new commit where the triggering issue was fixed. 🤝

adaptdl/adaptdl/torch/data.py
```diff
@@ -526,19 +490,19 @@ def __iter__(self):
             while not done:
                 self.sampler.set_epoch(
                     epoch, index=self._elastic.current_index)
-                self.batch_sampler.batch_size = self._elastic._sync_local_bsz()
+                self.batch_sampler.batch_size = self._elastic._context.get_batch_size()
```
odp, line 493: `_sync_local_bsz` cannot be replaced by the `Context.get_batch_size`, because `_sync_local_bsz` also does a broadcast from rank 0 to propagate the local batch size it calculated to rest of the replicas and also acts as barrier. `Context.get_batch_size` could cause local batch sizes in replicas to go out of sync.

VincentYaoMBZUAI: > `_sync_local_bsz` cannot be replaced by the `Context.get_batch_size`, because `_sync_local_bsz` also does a broadcast from rank 0 to propagate the local batch size it calculated to rest of the replicas and also acts as barrier. `Context.get_batch_size` could cause local batch sizes in replicas to go out of sync.

Thanks Omkar, please have a check on the new commit where the replacement issue is fixed.

